Issue #7, opened 2021-01-28
Ensure mypy_primer uses third party typeshed stubs in typeshed CI


hauntsaninja:
Following the refactor, this is no longer the case. Just filing this as a reminder to get things working.

JukkaL:
It would be good to have an option to install known third-party type packages generally, or at least in mypy CI. Since we no longer ship third-party package stubs with mypy, not having them limits the precision of type checking and inferred types, possibly quite significantly.

hauntsaninja:
Note that I've since updated most projects to install the type stubs they need from PyPI, so for the purposes of evaluating mypy regressions we're in good shape. But we still can't evaluate typeshed regressions on third party stubs (the core of this issue).

PR #9, opened 2021-04-24
add paroxython

Akuli:
When reviewing https://github.com/python/typeshed/pull/5248, I did a Github code search for projects that contain `import docopt  # type: ignore`. I had never heard of `paroxython` before, but I chose it because it's quite big and the latest commit is not too long ago.


hauntsaninja:
Thanks! Although note that mypy_primer currently doesn't work for third party package changes (broken by modular typeshed, https://github.com/hauntsaninja/mypy_primer/issues/7)

PR #12, opened 2021-06-12
feat: improve diff formatting

atakiar:
## Description

The goal of this PR is to improve the formatting of the diff reported - especially when it comes to multiline notes and errors. See python/mypy#10572 for an example of the current behavior

## Issue

Expected diff:

```shell
+ aioredis/client.py:4261: note:      Superclass:
+ aioredis/client.py:4261: note:          def (self: Redis) -> Any
+ aioredis/client.py:4261: note:      Subclass:
+ aioredis/client.py:4261: note:          def __aenter__(self) -> Coroutine[Any, Any, Pipeline]
```

Actual diff:

```shell
+ aioredis/client.py:4257: note:          def (self: Redis) -> Any
+ aioredis/client.py:4257: note:          def __aenter__(self) -> Coroutine[Any, Any, Pipeline]
+ aioredis/client.py:4257: note:      Subclass:
+ aioredis/client.py:4257: note:      Superclass:
```

## Potential Solution

To get the expected diff ordering, we can simply do away with sorting errors from the same line. However, this may result in a diff that indicates the same line was both added and removed, e.g.

```shell
- a.py:1: error ...
+ a.py:1: error ...
```

To handle this new issue we can filter these duplicates out by first identifying the exact type and number of removals and additions that are taking place

We then subtract these counters from one another simultaneously and get an appropriate count of removals and additions that must be present in our final diff

Finally, we iterate over the diff's lines, filtering with the counters, updating as we go, and report a diff in a more readable order

Let me know what you think!

### DIFF
--- a/mypy_primer.py
+++ b/mypy_primer.py
@@ -4,7 +4,6 @@ import argparse
 import asyncio
 import difflib
 import hashlib
-import itertools
 import multiprocessing
 import os
 import re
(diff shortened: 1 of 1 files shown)


hauntsaninja (commented):
Thanks, this is great. This also helps with mypy primer's use in typeshed, where introducing dependencies between modules can cause mypy to check files in a different order, also causing spurious diffs.

Two comments:
1) I think we no longer need the `itertools.groupby` at all, just the regex filter
2) The use of two counters makes this look complicated. Maybe the following is cleaner (and also helps test that I understand the logic):
```
net_change = defaultdict(int)
for line in diff_lines:
    net_change[line[2:]] += 1 if line[0] == "+" else -1  # eval(f"{line[0]}1") ;-)
return "\n".join(line for line in diff_lines if net_change[line[2:]])
```

atakiar:
Thanks for quick review!

1. Good catch, I've removed the `itertools.groupby`
2. Yeah that's a lot cleaner, but it's still missing something. Consider the following:
```python
old = """\
test.py:1: error: First
test.py:1: error: Second
test.py:1: error: Second
"""
new = """\
test.py:1: error: Second
test.py:1: error: First
"""
```
The net_change would be `defaultdict(<class 'int'>, {'test.py:1: error: Second': -1})` and we would print out:
```shell
+ test.py:1: error: Second
- test.py:1: error: Second
- test.py:1: error: Second
```
So, we need to handle both the sign and magnitude of the net_change, e.g. -1 means print out just 1 removal of that line. Also, once a line is processed, the net_change would need to be updated so we don't print it out a second time

hauntsaninja:
Thanks again!

PR #13, opened 2021-06-15
move porcupine out of failures expected section

Akuli:
Porcupine no longer bundles a custom typeshed in a git submodule.

### DIFF
--- a/mypy_primer.py
+++ b/mypy_primer.py
@@ -1236,6 +1236,10 @@ PROJECTS = [
         mypy_cmd="{mypy} paroxython",
         expected_success=True,
     ),
+    Project(
+        location="https://github.com/Akuli/porcupine.git",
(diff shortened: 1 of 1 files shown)


Akuli:
The CI was so quick that I don't think it actually checked much. Can this break something unexpectedly?

Akuli:
Nevermind. It really shouldn't break anything, since the success/failure expected sections are separated only by a comment :)

Akuli:
...except they are not, because I forgot to add `expected_success=True` :D

hauntsaninja:
Thanks! CI currently only runs lint. I'm lazy / it's slow / wasn't sure that there'd be much value in running mypy primer itself here / I'm hands on enough with typeshed that if things got catastrophically broken I'd notice and fix.

PR #14, opened 2021-06-23
add schema_salad & cwltool

mr-c:


mypy_primer.py
```diff
@@ -1241,6 +1241,18 @@ def inner() -> Optional[int]:
         mypy_cmd="{mypy} porcupine more_plugins",
         expected_success=True,
     ),
+    Project(
+        location="https://github.com/common-workflow-language/schema_salad.git",
+        mypy_cmd="MYPYPATH=$MYPYPATH:typeshed {mypy} --non-interactive "
```
JelleZijlstra, line 1246: This gave me pause because I like mypy-primer to use the real typeshed so we can test changes in it :) But it looks like this mostly adds additional third-party stubs?

mr-c: Correct, just additional 3rd party stubs

mr-c: If there is a better way to bundle the 3rd part stubs we need, I'm happy to change that!

hauntsaninja:
Thanks!

mr-c:
@hauntsaninja Thank you!

PR #15, opened 2021-06-29
Add edgedb

msullivan:


mypy_primer.py
```diff
@@ -1253,6 +1253,18 @@ def inner() -> Optional[int]:
         "--install-types cwltool/*.py tests/*.py",
         expected_success=True,
     ),
+    Project(
+        location="https://github.com/edgedb/edgedb.git",
+        mypy_cmd="{mypy} edb",
+        # weeeee, extract the deps by noping out setuptools.setup and reading them
```
JelleZijlstra, line 1259: lolwat

JelleZijlstra (approved):

mypy_primer.py
msullivan: Yeah, I dunno. I didn't want to install edgedb-server itself, because that requires compiling postgres and our rust CLI and a bunch of other stuff, and I don't want to duplicate things into a `requirements.txt` file, and pip doesn't have a way to say "install all the dependencies but don't build the package". I considered splitting out the dependency specification into another file, but that would still be hacky, so I decided to just go full hack.

hauntsaninja:
Great! Didn't realise plugins could be specified with file paths, that simplifies the setup.

PR #17, opened 2021-08-14
Create second project for mypy-protobuf

nipunn1313:
This was a rudimentary way of having multiple mypy commands
per project (as needed here). It might be more elegant to have
multiple mypy_cmd within a project, but it seemed nontrivial to
achieve, and I was able to whip this mechanism up more quickly -
using multiple projects with one mypy_cmd per project.
Not the most proud of it, but it works.


nipunn1313:
It's not super critical this gets in.
If the implementation strategy gives you the eebie jeebies, I think it's ok for this to wait for a better impl.

hauntsaninja:
Thanks, let me think about this one!

PR #19, opened 2021-09-03
Raise exception if selector matches nothing

nipunn1313:
I had issues w/ typos in my selector and ideally it would catch those
early and fails fast. This achieves that.

### DIFF
--- a/mypy_primer.py
+++ b/mypy_primer.py
@@ -530,11 +530,14 @@ class PrimerResult:
 def select_projects() -> Iterator[Project]:
     if ARGS.local_project:
         return iter([Project.from_location(ARGS.local_project)])
-    project_iter = (p for p in PROJECTS)
+    project_iter = iter(p for p in PROJECTS)
     if ARGS.project_selector:
-        project_iter = (
+        projects = [
             p for p in project_iter if re.search(ARGS.project_selector, p.location, flags=re.I)
-        )
+        ]
+        if projects == []:
+            raise Exception(f"No projects were selected by -k {ARGS.project_selector}")
+        project_iter = iter(projects)
     if ARGS.expected_success:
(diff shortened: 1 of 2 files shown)


test.sh
```diff
@@ -2,6 +2,6 @@ set -ex
 
 isort --diff --check --quiet .
 black --diff --check --quiet .
-flake8 --max-line-length 100 --ignore E203,W503
+flake8 --max-line-length 100 --ignore E203,W503 $(git ls-files | grep "py$")
```
nipunn1313, line 5: I had a .venv and it was hanging here because it would try to flake8 the entire venv.
Rolled in this fix here.

mypy_primer.py
```diff
@@ -527,14 +515,17 @@ def format_full(self) -> str:
 # ==============================
 
 
-def select_projects() -> Iterator[Project]:
+def select_projects() -> List[Project]:
```
hauntsaninja, line 518: The type annotation change here isn't correct, since we'll still return an iterator. (I'd be fine with actually returning a list too, if that's better for some reason)

hauntsaninja (changes requested):
Thanks, seems like a helpful change! :-)

mypy_primer.py
hauntsaninja: (I was surprised why mypy didn't complain, it's because we now just ignore project_iter)

nipunn1313: oh wow - this is actually messed up badly. It won't work if `-k` isn't provided. Unfortunate that it wasn't caught by mypy. Will manually test with and without `-k` and fix.

nipunn1313: Updated!

hauntsaninja:
Thanks, looks fixed!

PR #22, opened 2021-11-17
Add a repo with some tests for `pow` corner cases

AlexWaygood:


JelleZijlstra:
I wonder if we could actually put this in typeshed's own repo and point mypy-primer at it. That seems like a simple way to get testing in typeshed without having to maintain another repo.

AlexWaygood:
@JelleZijlstra I'd much prefer that,  if you and the other typeshed maintainers would be game!

AlexWaygood:
This obviously needs more discussion, so, closing this for now; sorry for the bother.

PR #26, opened 2022-01-06
Include python-sop

mr-c:
Would have caught this mypy 0.930 regression: https://ci.debian.net/data/autopkgtest/testing/amd64/p/python-sop/18100014/log.gz

```
sop/__init__.py:224: error: Missing type parameters for generic type "_SubParsersAction"
sop/__init__.py:387: error: Missing type parameters for generic type "_SubParsersAction"
```

### DIFF
--- a/mypy_primer.py
+++ b/mypy_primer.py
@@ -1520,7 +1520,7 @@ PROJECTS = [
         pip_cmd="{pip} install pytest pproxy",
     ),
     Project(
-        locaion="https://gitlab.com/dkg/python-sop",
+        location="https://gitlab.com/dkg/python-sop",
         mypy_cmd="{mypy} --strict sop",
     ),
 ]
--- a/mypy_primer.py
+++ b/mypy_primer.py
@@ -1521,7 +1521,8 @@ PROJECTS = [
     ),
     Project(
         locaion="https://gitlab.com/dkg/python-sop",
-        mypy_cmd="{mypy} --strict sop"
+        mypy_cmd="{mypy} --strict sop",
+    ),
 ]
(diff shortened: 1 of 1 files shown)


mypy_primer.py
```diff
@@ -1519,6 +1519,9 @@ def inner() -> Optional[int]:
         mypy_cmd="{mypy}",
         pip_cmd="{pip} install pytest pproxy",
     ),
+    Project(
+        locaion="https://gitlab.com/dkg/python-sop",
+        mypy_cmd="{mypy} --strict sop"
```
Akuli, line 1524:
```suggestion
        mypy_cmd="{mypy} --strict sop",
    ),
```

Akuli:
Regarding `_SubParsersAction`: I probably would have merged https://github.com/python/typeshed/pull/6178 even with the new mypy error. It's not uncommon for us to make classes generic to improve something, especially if the class is `_NamedLikeThis`. The proper solution to this would be default generics (https://github.com/python/typing/issues/307), but while that isn't available yet, you sometimes just have to change `Foo` to `Foo[Bar]` in type annotations.

mypy_primer.py
```diff
@@ -1519,6 +1519,10 @@ def inner() -> Optional[int]:
         mypy_cmd="{mypy}",
         pip_cmd="{pip} install pytest pproxy",
     ),
+    Project(
+        locaion="https://gitlab.com/dkg/python-sop",
```
Akuli, line 1523:
```suggestion
        location="https://gitlab.com/dkg/python-sop",
```

PR #31, opened 2022-05-06
Make third-party stubs work

### LINKED ISSUE #7: Ensure mypy_primer uses third party typeshed stubs in typeshed CI

Akuli:
Fixes #7. Tested in https://github.com/python/typeshed/pull/7786

### DIFF
--- a/mypy_primer.py
+++ b/mypy_primer.py
@@ -740,12 +740,7 @@ async def primer() -> int:
     )
 
     results = [
-        project.primer_result(
-            str(new_mypy),
-            str(old_mypy),
-            new_typeshed_dir,
-            old_typeshed_dir,
-        )
+        project.primer_result(str(new_mypy), str(old_mypy), new_typeshed_dir, old_typeshed_dir)
         for project in projects
     ]
     retcode = 0
--- a/mypy_primer.py
+++ b/mypy_primer.py
@@ -370,11 +370,20 @@ class Project:
         return mypy_cmd
 
     async def run_mypy(
-        self, mypy: Union[str, Path], additional_flags: Sequence[str] = ()
+        self,
+        mypy: Union[str, Path],
+        typeshed_dir: Path | None,
+        additional_flags: Sequence[str] = (),
     ) -> MypyResult:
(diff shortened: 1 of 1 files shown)


mypy_primer.py
```diff
@@ -369,12 +369,16 @@ def get_mypy_cmd(self, mypy: Union[str, Path], additional_flags: Sequence[str] =
...
-    ) -> MypyResult:
-        mypy_cmd = self.get_mypy_cmd(mypy, additional_flags)
+    async def run_mypy(self, mypy: Union[str, Path], typeshed_dir: Optional[Path]) -> MypyResult:
+        additional_flags = []
         env = os.environ.copy()
         env["MYPY_FORCE_COLOR"] = "1"
+
+        if typeshed_dir is not None:
+            additional_flags.append(f"--custom-typeshed-dir={typeshed_dir}")
+            env["MYPYPATH"] = ":".join(map(str, typeshed_dir.glob("stubs/*")))
```
hauntsaninja, line 379: Maybe append in the case where MYPYPATH is already set?

mypy_primer.py
```diff
@@ -633,6 +645,7 @@ async def inner(project: Project) -> Tuple[float, Project]:
 # ==============================
 
 
+# TODO: can't bisect issues with third-party stubs yet
```
hauntsaninja, line 648:
```suggestion
# TODO: can't bisect over typeshed commits yet
```

mypy_primer.py
```diff
@@ -383,21 +387,29 @@ async def run_mypy(
             cwd=ARGS.projects_dir / self.name,
             env=env,
         )
-        return MypyResult(
-            mypy_cmd, proc.stderr + proc.stdout, not bool(proc.returncode), self.expected_success
-        )
+
+        output = proc.stderr + proc.stdout
+        if typeshed_dir is not None:
```
hauntsaninja, line 392: (A little unfortunate... feels like we should be able to trick mypy into giving it the same treatment, I can look into this later)

hauntsaninja (commented):
Hooray!!!

Issue #8, opened 2021-03-11
How to add dependencies with plugins?


sobolevn:
Hi! Thanks a lot for this tool!

I am trying to add https://github.com/dry-python/returns support (because we have quite a lot of complex plugins).

Code:

```python
    Project(
        location="https://github.com/dry-python/returns.git",
        mypy_cmd="{mypy} returns",
        pip_cmd="{pip} install .",
        expected_success=True,
    ),
```

`returns` is based on `poetry`, not `pip`. But `pip` is able to install everything from `pyproject.toml` locally.

The next problem is that our plugin is not found:

```
old mypy
> /tmp/mypy_primer/old_mypy/venv/bin/mypy returns --python-executable=/tmp/mypy_primer/projects/_returns_venv/bin/python --no-incremental --cache-dir=/dev/null
UNEXPECTED FAILURE
        setup.cfg:153: error: Error importing plugin 'returns.contrib.mypy.returns_plugin': No module named 'returns'
        Found 1 error in 1 file (errors prevented further checking)
----------
```

Looks like the problem comes from a fact that `mypy` and `returns` use different venvs. But, I am not sure how to fix this.

hauntsaninja:
Yeah, mypy_primer doesn't yet support plugins — you'll notice mention of dry-python/returns at https://github.com/hauntsaninja/mypy_primer/blob/7c00c135fd2cacd2e08bd2fc81500d22b0fe3923/mypy_primer.py#L1249 :-) — and I'm not sure what the best way to support them is either.

The difficulty, as you mentioned, is that mypy and the projects are installed in different venvs. This allows us to reuse the same mypy venv across projects / easily change the mypy version independently.

I think one (possibly overkill) way of doing this is if we can get the project specific venv to "inherit" the mypy installation from the mypy venv. I bet there's some fun `.pth` hack way to do this.

sobolevn:
> you'll notice mention of dry-python/returns at 

@hauntsaninja yeah, this was a pleasant surprise! 🙂 

> I think one (possibly overkill) way of doing this is if we can get the project specific venv to "inherit" the mypy installation from the mypy venv. I bet there's some fun .pth hack way to do this.

I think that it is also possible to install `mypy` on top of the `returns` (for example) venv. Like `returns_env/bin/pip install /tmp/mypy_venv/path/to/mypy/setup.py`

sobolevn:
Yes, it works after a quick experiment:

```
[...]

PR #35, opened 2022-05-30
Add jsonschema libs

sirosen:
I'm adding more info to `types-jsonschema` stubs, and there were two bugs which it would have been nice to catch earlier. Add two users of `types-jsonschema` to the project list.

- [`check-jsonschema`](https://github.com/python-jsonschema/check-jsonschema) (my own work)
- [`synapse`](https://github.com/matrix-org/synapse), who reported and submitted a fix for a `types-jsonschema` bug

@DMRobertson, I don't believe that adding your project to mypy_primer carries any obligation for you or your team, but wanted to give you a heads-up about this.

We can also cut `synapse` and just add `check-jsonschema` if adding it is an issue for someone.

### DIFF
--- a/mypy_primer.py
+++ b/mypy_primer.py
@@ -1567,6 +1567,13 @@ PROJECTS = [
         mypy_cmd="{mypy} src",
         pip_cmd="{pip} install types-jsonschema types-requests",
     ),
+    Project(
+        location="https://github.com/matrix-org/synapse",
+        mypy_cmd="{mypy} synapse",
+        pip_cmd="{pip} install types-bleach types-commonmark types-jsonschema "
+        "types-opentracing types-Pillow types-psycopg2 types-pyOpenSSL types-PyYAML "
+        "types-requests types-setuptools"
+    )
 ]
 assert len(PROJECTS) == len({p.name for p in PROJECTS})
 
--- a/mypy_primer.py
+++ b/mypy_primer.py
@@ -1567,13 +1567,6 @@ PROJECTS = [
         mypy_cmd="{mypy} src",
         pip_cmd="{pip} install types-jsonschema types-requests",
     ),
-    Project(
-        location="https://github.com/matrix-org/synapse",
-        mypy_cmd="{mypy} synapse",
-        pip_cmd="{pip} install types-bleach types-commonmark types-jsonschema "
-        "types-opentracing types-Pillow types-psycopg2 types-pyOpenSSL types-PyYAML "
-        "types-requests types-setuptools",
-    ),
 ]
 assert len(PROJECTS) == len({p.name for p in PROJECTS})
 
--- a/mypy_primer.py
+++ b/mypy_primer.py
@@ -1372,6 +1372,7 @@ PROJECTS = [
             " @typechecked-files --no-strict-optional"
         ),
         pip_cmd="{pip} install types-PyYAML",
+        expected_success=True,
     ),
     Project(
         location="https://github.com/wntrblm/nox",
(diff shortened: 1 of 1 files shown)


DMRobertson:
> @DMRobertson, I don't believe that adding your project to mypy_primer carries any obligation for you or your team, but wanted to give you a heads-up about this.
> 
> We can also cut `synapse` and just add `check-jsonschema` if adding it is an issue for someone.

Thanks for the heads-up! No objections from us, but I would like to sound a few notes of caution:
* Synapse isn't completely type-checked. Last time I got mypy to generate a report I think around 10% of expressions in type-checked modules were Any.
* Synapse uses two mypy plugins: https://github.com/Shoobx/mypy-zope and [our own plugin](https://github.com/matrix-org/synapse/blob/develop/scripts-dev/mypy_synapse_plugin.py). I don't know if that causes pain for `mypy_primer`.
* We run our typechecking against pinned versions of stubs, packages and mypy itself. There is a daily CI job which tries against the latest version of all of these so we get some warning of breakages. I try to keep on top of those if they break (that's how we found https://github.com/python/typeshed/pull/7980), but we don't promise to always typecheck on the latest version of everything!
* We might pull in new type stubs in the future---I'm not sure if the `pip_cmd` could get out of date?
* Lastly: we don't use `jsonschema` a huge amount---but then again, we use it enough for us to have spotted https://github.com/python/typeshed/pull/7980.

mypy_primer.py
```diff
@@ -1567,6 +1567,13 @@ def inner() -> Optional[int]:
         mypy_cmd="{mypy} src",
         pip_cmd="{pip} install types-jsonschema types-requests",
     ),
+    Project(
+        location="https://github.com/matrix-org/synapse",
+        mypy_cmd="{mypy} synapse",
+        pip_cmd="{pip} install types-bleach types-commonmark types-jsonschema "
+        "types-opentracing types-Pillow types-psycopg2 types-pyOpenSSL types-PyYAML "
+        "types-requests types-setuptools"
+    )
```
DMRobertson, line 1576: I think `black` wants trailing commas here?

```suggestion
        "types-requests types-setuptools",
    ),
```

sirosen: Yep, thanks for the fix!

sirosen:
I think that it's okay that not everything inside the library is annotated, so long as it passes `mypy` without errors. A bunch of `Any`s is okay; failing `mypy` probably isn't (though I don't know for sure).

More importantly, I saw that you're _using_ a bunch of different stubs from typeshed. IMO, that makes the project a sort of attractive option to include, since it covers many different parts of typeshed, even if it only uses each one a little.

> * Synapse uses two mypy plugins: https://github.com/Shoobx/mypy-zope and [our own plugin](https://github.com/matrix-org/synapse/blob/develop/scripts-dev/mypy_synapse_plugin.py). I don't know if that causes pain for `mypy_primer`.

This seems to me like the biggest likely issue. I'll wait to hear from maintainers about this, but I probably wouldn't have suggested the integration if I'd known that there's an internal plugin. The internal plugin in particular could make `mypy_primer` overly sensitive to changes in that project.

Everything else seems pretty much okay to me. I see a commented out block where one project was listed as pulling the `types-*` packages out of a `setup.py`, but everything else seems vulnerable to the issue of getting out of date on needed stubs.

JelleZijlstra:
See #8 for discussion of plugins. But I think it's fine if mypy-primer projects get some mypy errors: what matters is that they give us useful information if something changes.

sirosen:
I removed `synapse` since it sounds like it can't be added (yet), so this PR is down to just `check-jsonschema` for now.

I've started looking for other projects using `types-jsonschema`, but I'll save any additions for a future change.

hauntsaninja:
Thanks! I will one day get around to adding support for plugins :-)

PR #39, opened 2022-06-13
Add canonical's cloud-init to projects to check

sirosen:
The mypy invocation is taken from their tox config, as is the list of typeshed stubs to use.

cc @holmanb , this might interest you as I see you're the @canonical team member who most recently made changes to the mypy config for `cloud-init`. Being in mypy_primer doesn't carry any obligation for you, but it means your project will be used to examine typeshed and mypy changes to assess the impact.
At least, I don't think there's any obligation -- someone please correct me if I'm wrong!

### DIFF
--- a/mypy_primer.py
+++ b/mypy_primer.py
@@ -1569,6 +1569,13 @@ PROJECTS = [
         mypy_cmd="{mypy} discord",
         pip_cmd="{pip} install types-requests types-setuptools aiohttp",
     ),
+    Project(
+        location="https://github.com/canonical/cloud-init",
+        mypy_cmd="{mypy} cloudinit/ tests/ tools/",
+        pip_cmd="{pip} install jinja2 pytest "
+        "types-jsonschema types-oauthlib "
+        "types-pyyaml types-requests types-setuptools",
+    ),
 ]
 assert len(PROJECTS) == len({p.name for p in PROJECTS})
 


sirosen:
Oh, I forgot to mention my motivation: I plan to be doing a bunch of work on `types-jsonschema` over the coming year, and want `mypy_primer` to have a couple more projects which use those stubs, so that the chances of introducing bugs are lower. `cloud-init` seems to me like a really good candidate.

hauntsaninja:
Nope, no obligation for anyone else!

I'd also be willing to add https://github.com/matrix-org/synapse even without the plugin, as long as type checking it does something reasonable without the plugin.

holmanb:
@sirosen Hi there! Thanks for the ping.

No objections, sounds good to me!

A couple of notes, in case it helps:


- Our adoption of mypy was fairly recent (Feb 2022)
- Our adoption of type annotation was fairly limited up until Jan 2022, when we dropped 3.5 support. Plenty of additions (and bugs fixed) since then.
- The codebase [currently provides support for 3.6+](https://cloudinit.readthedocs.io/en/latest/topics/contributing.html?highlight=python%20version#python-support)

sirosen:
> I'd also be willing to add https://github.com/matrix-org/synapse even without the plugin, as long as type checking it does something reasonable without the plugin.

I'm happy to circle back on this at some later date... but I'm also okay not worrying about it.

I'm still casually looking at https://github.com/search?q=types-jsonschema&type=code to see if I can find anyone else who should be included for my (selfish!) purposes. Suggestions of other ways of finding prominent typeshed users very much welcome. 😄 

---

Any objection to merging this as it currently stands? It seems like we're all happy to add cloud-init to mypy_primer!

holmanb:
@sirosen Sure, go for it! I don't think you even need to ask any of the devs, but the ping / interaction is welcome. You can always drop a line in irc too if questions crop up.

hauntsaninja (approved):
Thanks!

PR #41, opened 2022-07-16
Optionally run mypyc on projects

zsol:


hauntsaninja (commented):
Thanks for working on this! Left two quick comments.

Do you have a sense of how long this takes? Might need to be able to flag this off to avoid slowing down typeshed CI (which won't benefit from this)

zsol:
For Black this takes around 6 minutes on my laptop. Definitely worth gating off for typeshed. I can add a command line flag if that helps

hauntsaninja:
A command line flag would be great!

zsol:
I've addressed your comments in the new commit, going to add a command line flag now. Thanks for the blazing fast reviews :)

mypy_primer.py
```diff
@@ -519,11 +552,32 @@ def _get_diff(self) -> str:
...
+                    if verbose:
+                        mypyc_result += f"\n{self.new_mypyc_result.output}"
+            else:
+                mypyc_result += "✅]"
+        return mypyc_result
+
     def format_concise(self) -> str:
-        if self.diff:
-            return f"{self.project.name} ({self.project.location})\n{self.diff}"
+        mypyc_result = self.mypyc_result(verbose=False)
```
hauntsaninja, line 578: mypyc_result will always be truthy for black. We should want this check to only be true if there is a change in mypyc_result (or also if it's broken)

zsol: Hmm do we want to show results if old is broken but new is fixed?

zsol: I assumed not in the latest commit

hauntsaninja: I think we do want to show the change; it's useful to confirm that you've fixed a thing. This is consistent with mypy_primer always showing a diff if there is a diff, even if that diff is "good"

zsol: I've addressed this in the latest commit

mypy_primer.py
```diff
@@ -964,6 +1035,11 @@ def inner() -> Optional[int]:
         location="https://github.com/psf/black",
         mypy_cmd="{mypy} src",
         pip_cmd="{pip} install types-dataclasses types-typed-ast",
+        mypyc_cmd=(
+            "{python} setup.py --use-mypyc install "
+            "&& {python} -m pip install pytest"
+            "&& {python} -m pytest -k 'not incompatible_with_mypyc' tests/test_black.py"
```
graingert, line 1041: Instead of passing a bespoke set of flags here, mypy should configure an ini option in `conftest.py` so you can run
```suggestion
            "&& {bin}/pytest --mypy-primer
```

zsol: I'm not convinced the right idea is to run Black's own test suite here, maybe better to run `{python} -m black check .`

mr-c:
Ping; it would be great to have this for schema-salad and cwltool given the mypyc regressions we experienced in the last two releases

mr-c:
Hello from 2024, can someone take this over?

PR #44, opened 2022-09-16
Update bokeh invocation

AlexWaygood:
`bokeh` depends on `selenium`, and `selenium` has shipped with inline type annotations since release 4.1.2

### DIFF
--- a/mypy_primer.py
+++ b/mypy_primer.py
@@ -1647,7 +1647,7 @@ PROJECTS = [
     Project(
         location="https://github.com/bokeh/bokeh",
         mypy_cmd="{mypy} src release",
-        pip_cmd="{pip} install types-boto tornado numpy jinja2",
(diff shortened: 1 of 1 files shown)


srittau (approved):

srittau:
Thanks!

mypy_primer.py
```diff
@@ -1647,7 +1647,7 @@ def inner() -> Optional[int]:
     Project(
         location="https://github.com/bokeh/bokeh",
         mypy_cmd="{mypy} src release",
-        pip_cmd="{pip} install types-boto tornado numpy jinja2",
+        pip_cmd="{pip} install types-boto tornado numpy jinja2 selenium",
```
Akuli, line 1650: I wonder if `types-boto` is still needed, because we now set `MYPYPATH` so that it contains all third-party stubs.

AlexWaygood: Yeah, probably not!

PR #47, opened 2022-09-29
Add nionutils project (strict typing).

cmeyer:


cmeyer:
I'm not sure this is ready for merging yet - I thought the PR would run the tests.

The question is whether the mypy command runs from within the cloned directory or from the directory above?

mypy_primer.py
```diff
@@ -1642,6 +1642,10 @@ def inner() -> int | None:
         location="https://github.com/pandas-dev/pandas-stubs",
         mypy_cmd="{mypy} pandas-stubs tests",
     ),
+    Project(
+        location="https://github.com/nion-software/nionutils",
+        mypy_cmd="{mypy} --strict -p nion.utils",
```
hauntsaninja, line 1647: This will be the default in the next mypy release, but I think you need this for now
```suggestion
        mypy_cmd="{mypy} --namespace-packages --strict -p nion.utils",
```

hauntsaninja (commented):
Thanks! Ran it locally and had one comment

hauntsaninja:
It runs within the cloned directory. mypy_primer CI doesn't run mypy_primer, because you'd effectively need a mypy_primer_primer to get reasonable output :-)

mypy_primer.py
cmeyer: Yes, I dropped that at the last second since this is my lowest level library. I'll resubmit.

Btw, once I work out the 0.981, I'll have several more projects to add, all strictly typed in namespaces.

cmeyer:
Ok this should be ready to go.

hauntsaninja (approved):

hauntsaninja:
Thanks!

PR #50, opened 2022-11-08
Add --additional-flags argument

JelleZijlstra:
So we can use it in typeshed to disable bytearray/memoryview promotions


hauntsaninja (commented):
Does this work? Won't you have to either stick hyphens back on or use argparse.REMAINDER or something?

JelleZijlstra:
Yes, I tested locally with like `--additional-flags="--skip-bytearray-promotion --skip-memoryview-promotion"`.

hauntsaninja:
Oh right, I forgot about equals. Thanks!

PR #54, opened 2022-11-19
Replace types-attrs with attrs

cdce8p:
~Blocked by https://github.com/python/mypy/issues/14137~

### DIFF
--- a/mypy_primer.py
+++ b/mypy_primer.py
@@ -977,7 +977,7 @@ PROJECTS = [
     Project(
         location="https://github.com/pytest-dev/pytest",
         mypy_cmd="{mypy} src testing",
-        pip_cmd="{pip} install py types-attrs types-setuptools",
+        pip_cmd="{pip} install attrs py types-setuptools",
         expected_success=True,
     ),
     Project(
@@ -1042,7 +1042,7 @@ PROJECTS = [
     Project(
         location="https://github.com/scrapy/scrapy",
         mypy_cmd="{mypy} scrapy tests",
-        pip_cmd="{pip} install types-attrs types-pyOpenSSL types-setuptools",
+        pip_cmd="{pip} install attrs types-pyOpenSSL types-setuptools",
         expected_success=True,
     ),
     Project(
@@ -1209,7 +1209,7 @@ PROJECTS = [
     Project(
         location="https://github.com/Textualize/rich",
         mypy_cmd="{mypy} -p rich --ignore-missing-imports --warn-unreachable",
-        pip_cmd="{pip} install types-attrs",
+        pip_cmd="{pip} install attrs",
         expected_success=True,
     ),
     Project(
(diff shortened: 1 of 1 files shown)


hauntsaninja:
We already install types-attrs. I think the bug is a cache bug, so it won't repro in mypy_primer (which only does from scratch checking). Or at least, I wasn't able to repro with `mypy_primer -k home-assistant` — are you able to?

AlexWaygood:
It might be worth removing the `types-attrs` import and replacing it with `attrs` regardless of the mypy bug, since `types-attrs` was deleted from typeshed in June 2021 after `attrs` introduced inline types, and `home-assistant` no longer appears to use `types-attrs`.

cdce8p:
> We already install types-attrs.

🤦🏻 I should have looked more carefully. I just assumed that it wasn't installed since I was able to crash mypy with just `attrs` or `types-attrs` installed and the primer didn't catch it.

> I think the bug is a cache bug, so it won't repro in mypy_primer (which only does from scratch checking). Or at least, I wasn't able to repro with `mypy_primer -k home-assistant` — are you able to?

It is a cache bug but one that occurs on the first write / serialize. I'm able to reproduce it with `--no-incremental`. Interestingly though setting `--cache-dir=/dev/null` (like the primer does) hides it 🤔

> It might be worth removing the `types-attrs` import and replacing it with `attrs`

Will do that.

hauntsaninja:
It looks like there are several projects that use types-attrs instead of attrs, want to upgrade all of them?

cdce8p:
> > I think the bug is a cache bug, so it won't repro in mypy_primer (which only does from scratch checking). Or at least, I wasn't able to repro with `mypy_primer -k home-assistant` — are you able to?
> 
> It is a cache bug but one that occurs on the first write / serialize. I'm able to reproduce it with `--no-incremental`. Interestingly though setting `--cache-dir=/dev/null` (like the primer does) hides it 🤔

Does it make sense to look for ways to remove the `--cache-dir=/dev/null` argument? Maybe if we replace it with `--no-incremental`?

hauntsaninja (approved):
Thank you, this looks good!

Good question. primer sets both `--no-incremental` and `--cache-dir=/dev/null`. I guess we could set some `--cache-dir`, we might just want to ensure that all the different concurrent mypys write to different caches. The cost of dropping `--cache-dir=/dev/null` is basically figuring that out, some extra CI slowness, and some personal dev workflows that I can adjust.

I do confess to being surprised at the existence of an issue this would have caught that wasn't caught by the regular mypy self check. So updating my priors, but still don't think it's critical if we don't do it :-)

cdce8p:
> It is a cache bug but one that occurs on the first write / serialize. I'm able to reproduce it with `--no-incremental`. Interestingly though setting `--cache-dir=/dev/null` (like the primer does) hides it 🤔

I've opened a PR to run serialize even with `--cache-dir=/dev/null`. Might be useful to have.
https://github.com/python/mypy/pull/14155

PR #55, opened 2022-12-01
Allow primer to run on Windows

Avasam:
Fixes the following issue: `FileNotFoundError: [WinError 3] The system cannot find the path specified: 'E:\\tmp\\mypy_primer'`
By using the OS' temp folder.

Fixes the mypy.exe expected location (from `/bin/mypy` to `/scripts/mypy.exe` on windows)

Ensure utf-8 encoding

FIx `--soft-error-limit ' -1'` becomming `--soft-error-limit \' -1\'` error in powershell

Fixed windows backward slashes in path resulting in bad escape regex errors.

### DIFF
--- a/mypy_primer.py
+++ b/mypy_primer.py
@@ -898,7 +898,7 @@ def parse_options(argv: list[str]) -> argparse.Namespace:
     primer_group.add_argument("--debug", action="store_true", help="print commands as they run")
     primer_group.add_argument(
         "--base-dir",
-        default=Path(tempfile.gettempdir()) / "mypy_primer",
+        default=Path(tempfile.gettempdir() if sys.platform == "win32" else "/tmp") / "mypy_primer",
         type=Path,
         help="dir to store repos and venvs",
     )
--- a/mypy_primer.py
+++ b/mypy_primer.py
@@ -25,7 +25,7 @@ from typing import Any, Awaitable, Callable, Iterator, Sequence, TypeVar, Union
 
 T = TypeVar("T")
 RevisionLike = Union[str, None, Callable[[Path], Awaitable[str]]]
-
+BIN_DIR = "scripts" if sys.platform == "win32" else "bin"
 
 # ==============================
 # utils
@@ -219,7 +219,7 @@ async def setup_mypy(mypy_dir: Path, revision_like: RevisionLike, editable: bool
     mypy_dir.mkdir(exist_ok=True)
     venv_dir = mypy_dir / "venv"
     venv.create(venv_dir, with_pip=True, clear=True)
-    pip_exe = str(venv_dir / "bin" / "pip")
+    pip_exe = str(venv_dir / BIN_DIR / "pip")
 
     if ARGS.mypyc_compile_level is not None:
         editable = True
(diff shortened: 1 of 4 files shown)


mypy_primer.py
```diff
@@ -443,7 +444,7 @@ async def source_paths(self, mypy_python: str) -> list[Path]:
     def from_location(cls, location: str) -> Project:
         additional_flags = ""
         if Path(location).is_file():
-            with open(location) as f:
+            with open(location, encoding="UTF-8") as f:
```
hauntsaninja, line 447: This one should probably use default system encoding? Most common use case is just typing stuff up a bisection case in an editor

Avasam: It may not be necessary. But still good practice to specify the encoding until the default in python-windows is changed from the system default (Windows-1252) to "utf-8" (same as linux and darwin).

Ref: https://peps.python.org/pep-0597/

hauntsaninja (approved):
Great, thanks for making this work! I changed two quick things: 1) continue to use /tmp on unix because I often go to that dir manually, 2) use re.escape for the regex.

mypy_primer.py
```diff
@@ -898,7 +894,7 @@ def parse_options(argv: list[str]) -> argparse.Namespace:
     primer_group.add_argument("--debug", action="store_true", help="print commands as they run")
     primer_group.add_argument(
         "--base-dir",
-        default=Path(tempfile.gettempdir()) / "mypy_primer",
+        default=Path(tempfile.gettempdir() if sys.platform == "win32" else "/tmp") / "mypy_primer",
```
Avasam, line 897: Isn't /tmp the temp dir on linux? I meant this code to be os-agnostic

mypy_primer.py
```diff
@@ -492,12 +492,7 @@ def _get_diff(self) -> str:
         new_lines = new_output.splitlines()
         # Hide "note" lines which contain ARGS.base_dir... this hides differences between
         # file paths, e.g., when mypy points to a stub definition.
-
-        if sys.platform == "win32":
-            base_dir = str(ARGS.base_dir).replace("\\", r"\\")
-        else:
-            base_dir = ARGS.base_dir
-        base_dir_re = re.compile(f"{base_dir}.*: note:")
+        base_dir_re = re.compile(f"{re.escape(ARGS.base_dir)}.*: note:")
```
Avasam, line 495: Ah nice. I was looking for something like this :)

mypy_primer.py
hauntsaninja: /tmp also seems to work well enough on macOS... It's probably not the best temporary location, but `tempfile.gettempdir()` gives you stuff like `/var/folders/0y/wt5dx2vj6t16wq97pglrkynm0000gp/T` and that's a pain to mess with manually

Ezekiel-create (commented):
z

mypy_primer.py
```diff
@@ -241,7 +242,7 @@ async def setup_mypy(mypy_dir: Path, revision_like: RevisionLike, editable: bool
         if ARGS.mypyc_compile_level is not None:
             env = os.environ.copy()
             env["MYPYC_OPT_LEVEL"] = str(ARGS.mypyc_compile_level)
-            python_exe = str(venv_dir / "bin" / "python")
```
Ezekiel-create, line 244: tmp

PR #56, opened 2022-12-01
typed ARGS

Avasam:
I was a bit confused about the expected type of certain parameters when debugging #55 . So I typed `ARGS` for ease of use.

### DIFF
--- a/mypy_primer.py
+++ b/mypy_primer.py
@@ -585,7 +585,7 @@ def select_projects() -> Iterator[Project]:
 # hidden entrypoint logic
 # ==============================
 
-RECENT_MYPYS = ["0.971", "0.961"]
+RECENT_MYPYS = ["0.991", "0.982", "0.971", "0.961"]
 
 
 async def validate_expected_success() -> None:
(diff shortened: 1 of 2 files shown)


hauntsaninja (changes requested):
This isn't particularly type safe, since `argparse.Namespace` has `__getattr__`, e.g. you're missing several attributes from Args. If we want to make something more type safe, I'd make a dataclass and then do something like `ret = Args(**vars(parser.parse_args(argv)))`

mypy_primer.py
```diff
@@ -903,13 +903,43 @@ def parse_options(argv: list[str]) -> argparse.Namespace:
...
+    old: str | None
+    old_success: bool
+    old_typeshed: str | None
+    output: str
+    project_date: str | None
+    project_selector: str | None
+    repo: str
+    shard_index: int | None
+    validate_expected_success: bool
+    projects_dir: Path = None  # type: ignore[assignment]  # Initialized later
```
hauntsaninja, line 939:
```suggestion
    projects_dir: Path = field(init=False)
```

hauntsaninja (commented):
Thanks, looks great! Maybe we can use an init=False field for projects_dir?

hauntsaninja (approved):
Thank you!

PR #57, opened 2022-12-01
Fix a couple more Windows incompatibilities

AlexWaygood:
On Windows, MYPYPATH needs to be a semicolon-delimited list rather than a colon-delimited list


hauntsaninja (approved):
Thanks!

PR #60, opened 2022-12-03
Shard projects more evenly

### LINKED ISSUE #58: Better balanced shards
I would *love* if mypy primer better balanced sharding. On a recent PR to mypy, I noticed that:
 - shard 0 only took 7 minutes
 - shard 1 took 27 minutes
 - shard 2 took 14 minutes
 - shard 3 took 20 minutes
 - shard 4 took 21 minutes

(Note that mypy-primer could take ~10 minutes less if it optimally balanced)

I know that it would be infeasible to construct lists for every single combination, so I propose:

What if every project had a "difficulty" number that was a rough estimate of time mypy takes to type check it? The idea is that you could try to balance these numbers into buckets (just use a greedy approach: from largest difficulty to smallest just always put it in the lowest difficulty bucket).

I'm not sure how we could keep these up to date though. Is there a metric that is simple to take but that correlates with mypy runtime? Number of files? Dependencies? Lines of code? Count of `import typing`?

hauntsaninja:
Fixes #58

### DIFF
--- a/mypy_primer.py
+++ b/mypy_primer.py
@@ -321,7 +321,7 @@ class Project:
     # if expected_success, there is a recent version of mypy which passes cleanly
     expected_success: bool = False
     # cost is vaguely proportional to type check time
-    cost: float = 3
+    cost: int = 3
 
     @property
     def name(self) -> str:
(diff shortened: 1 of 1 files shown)


hauntsaninja:
cc @A5rocks

A5rocks:
Looks good -- assuming the code for distributing into buckets works as expects 😅

hauntsaninja:
It looks like it works, but my first attempt didn't, in a somewhat interesting way.

Originally I had `shard_projects= [[]] * ARGS.num_shards)` instead of `shard_projects = [[] for _ in range(ARGS.num_shards)]`. In the first you end up with multiple references to the same list

PR #61, opened 2022-12-03
Fix escape code rendering on Windows + Update typeshed main branch in doc

Avasam:
Demo:
image
\+ centralized the platform-specific constants under a single check.

\+ Updated the `strip_colour_code` regex:
- Standardize `\033` vs `\x1b`
- Regex matches codes more tightly

I've also updated the name of the default branch for typeshed in the doc cuz I noticed that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ mypy_primer -o diff
 
 See the impact of your risky change with:
 ```
-mypy_primer --repo https://github.com/hauntsaninja/mypy.git --new my_risky_change --old main
+mypy_primer --repo https://github.com/hauntsaninja/mypy.git --new my_risky_change --old master
 ```
 
 See the impact of your risky typeshed change with:
--- a/README.md
+++ b/README.md
@@ -107,12 +107,12 @@ mypy_primer -o diff
 
 See the impact of your risky change with:
 ```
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -107,12 +107,12 @@ mypy_primer -o diff
 
 See the impact of your risky change with:
 ```
-mypy_primer --repo https://github.com/hauntsaninja/mypy.git --new my_risky_change --old master
+mypy_primer --repo https://github.com/hauntsaninja/mypy.git --new my_risky_change --old main
```
hauntsaninja, line 110: mypy still uses master

mypy_primer.py
```diff
@@ -1713,6 +1726,37 @@ def inner() -> int | None:
         mypy_cmd="{mypy} src",
         pip_cmd="{pip} install pydantic beartype hydra-core",
     ),
+    Project(
```
hauntsaninja, line 1729: Want to add new projects in a separate PR?

Avasam: I meant to unstage that. I accidentally pushed with my second commit

hauntsaninja (approved):

hauntsaninja:
Thank you!

PR #63, opened 2022-12-03
Fix sharding

### LINKED ISSUE #62: Significant recent performance regression?
mypy_primer jobs seem to be taking 30+ minutes (or more) on typeshed PRs today. The norm previously was around ~12-13 minutes.

https://github.com/python/typeshed/actions/workflows/mypy_primer.yml

JelleZijlstra:
Fixes #62

### DIFF
--- a/mypy_primer.py
+++ b/mypy_primer.py
@@ -587,7 +587,8 @@ def select_projects() -> list[Project]:
     if projects == []:
         raise ValueError("No projects selected!")
 
-    if ARGS.num_shards and ARGS.shard_index is not None:
+    if ARGS.num_shards:
+        assert ARGS.shard_index is not None
         shard_costs = [0] * ARGS.num_shards
         shard_projects: list[list[Project]] = [[] for _ in range(ARGS.num_shards)]
         for p in sorted(projects, key=lambda p: (p.cost, p.location), reverse=True):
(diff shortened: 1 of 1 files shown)


mypy_primer.py
```diff
@@ -587,7 +587,7 @@ def select_projects() -> list[Project]:
     if projects == []:
         raise ValueError("No projects selected!")
 
-    if ARGS.num_shards and ARGS.shard_index:
+    if ARGS.num_shards and ARGS.shard_index is not None:
```
AlexWaygood, line 590: Hmm, surely `ARGS.shard_index` should never be `None` if `ARGS.num_shards` is not `None`. This is guarded against here:

https://github.com/hauntsaninja/mypy_primer/blob/592a1f7972989e967da17bbbfb851e27aea9202a/mypy_primer.py#L910

Maybe using an `assert` here would be better?

```suggestion
    if ARGS.num_shards:
        assert ARGS.shard_index is not None
```

AlexWaygood (approved):

AlexWaygood:
I triggered a rerun of the primer job for the pywin32 typeshed PR, to double-check that this fixes the issue: https://github.com/python/typeshed/actions/runs/3608748067

AlexWaygood:
Looks good — we're back to sane mypy_primer performance, and the `comtypes` error has been deduplicated: https://github.com/python/typeshed/pull/9308#issuecomment-1336197547 🎉

hauntsaninja:
Thanks for fixing! I added the guard to shut up the type checker after I had tested the change

Issue #43, opened 2022-08-05
add more projects that use pyright


hauntsaninja:
The couple ones we already have have been disproportionately useful for implementing new typing features in mypy

https://github.com/awslabs/gluon-ts
https://github.com/canonical/operator
https://github.com/astropenguin/xarray-dataclasses
https://github.com/Rapptz/RoboDanny

A5rocks:
Given `mypy_primer` now is used as a pyright primer (ish) too, I suspect this issue could probably be closed? Or maybe pushed off to, like, pyright's maintainer at least! They know best!

Issue #42, opened 2022-07-21
add more projects to mypy_primer


hauntsaninja:
https://github.com/pybind/pybind11
https://github.com/ronf/asyncssh
https://github.com/cytopia/pwncat
https://github.com/NervanaSystems/ngraph
https://github.com/openvinotoolkit/openvino/blob/master/src/bindings/python/setup.cfg
https://github.com/nv-legate/cunumeric
https://github.com/geofront-auth/geofront
https://github.com/whonore/Coqtail
https://github.com/relastle/glacier
https://github.com/qutebrowser/qutebrowser
https://github.com/pola-rs/polars
https://github.com/lark-parser/lark
https://github.com/tomerfiliba/plumbum
https://github.com/PyCQA/astroid
https://github.com/PyCQA/flake8-pyi
https://github.com/scikit-build/scikit-build
https://github.com/pypa/build
https://github.com/pypa/readme_renderer
https://github.com/xaviml/controllerx
https://github.com/bimmerconnected/bimmer_connected
https://github.com/scikit-hep/particle
https://github.com/scikit-hep/vector

hauntsaninja:
https://github.com/mongodb/mongo-python-driver reported a couple typeshed regressions

hauntsaninja:
https://github.com/artigraph/artigraph could have caught a mypy regression

hauntsaninja:
https://github.com/enthought/comtypes

hauntsaninja:
https://github.com/trezor/trezor-firmware once next mypy release is out

AlexWaygood:
Sounds like https://github.com/caronc/appraise might have caught https://github.com/python/mypy/issues/13536

hauntsaninja:
Maybe some of the projects that are now referring to https://github.com/python/mypy/issues/731

hauntsaninja:
https://github.com/scikit-hep/pyhf would catch https://github.com/python/mypy/issues/13813

Avasam:
This isn't a request to add any of 'em, rather just another metric/datapoint to be aware of some of mypy's big users. Using the https://github.com/github-tooling/ghtopdep CLI tool, I listed the top 100 packages and top 100 repositories (by star count) that have mypy as a dependent on Github (because Github doesn't allow to sort "dependants"):

<details>
<summary><code>ghtopdep https://github.com/python/mypy --packages --rows 100</code></summary>
<br/>

| url                                                          | stars   | primer           |
[...]

PR #65, opened 2022-12-05
Added my open source projects

Avasam:
All for are checked with Pyright (all strict except `speedrun.com_global_scoreboard_webapp`, I'm due for my bi-yearly update :P ) #43

The top two are libraries checked with mypy as well. #42

The bottom two still have a mypy config file for support, but I don't actively lint against it because mypy has too many annoying quirks.

### DIFF
--- a/mypy_primer.py
+++ b/mypy_primer.py
@@ -1499,6 +1499,18 @@ PROJECTS = [
         pip_cmd="{pip} install types-pyflakes",
         expected_success=True,
     ),
+    Project(
+        location="https://github.com/JohannesBuchner/imagehash",
+        mypy_cmd="{mypy} imagehash",
+        pip_cmd="{pip} install -r requirements-conda.txt types-Pillow",
+        expected_success=True,
+    ),
+    Project(
+        location="https://github.com/Kalmat/PyWinCtl",
+        mypy_cmd="{mypy} src/pywinctl",
+        pip_cmd="{pip} install -r docs/requirements.txt",
+        expected_success=True,
+    ),
     *(
         [
             Project(
@@ -1744,6 +1756,19 @@ PROJECTS = [
         mypy_cmd="{mypy} src",
         pip_cmd="{pip} install pydantic beartype hydra-core",
         cost=30,
+    ),
+    Project(
+        location="https://github.com/Avasam/Auto-Split",
+        mypy_cmd="{mypy} src",
+        pip_cmd="{pip} install -r scripts/requirements.txt types-d3dshot types-keyboard"
+        " types-Pillow types-psutil types-PyAutoGUI types-pyinstaller types-pywin32 types-requests"
+        " types-toml",
+    ),
+    Project(
+        location="https://github.com/Avasam/speedrun.com_global_scoreboard_webapp",
+        mypy_cmd="{mypy} backend",
+        pip_cmd="{pip} install -r scripts/requirements.txt types-Flask types-Flask-SQLAlchemy"
+        " types-httplib2 types-requests",
     ),
 ]
 assert len(PROJECTS) == len({p.name for p in PROJECTS})


mypy_primer.py
```diff
@@ -1745,6 +1757,19 @@ def inner() -> int | None:
...
+        location="https://github.com/Avasam/Auto-Split",
+        mypy_cmd="{mypy} src",
+        pip_cmd="{pip} install -r scripts/requirements.txt types-d3dshot types-keyboard"
+        " types-Pillow types-psutil types-PyAutoGUI types-pyinstaller types-pywin32 types-requests"
+        " types-toml",
+    ),
+    Project(
+        location="https://github.com/Avasam/speedrun.com_global_scoreboard_webapp",
+        mypy_cmd="{mypy} backend",
+        pip_cmd="{pip} install -r scripts/requirements.txt types-Flask types-Flask-SQLAlchemy"
```
hauntsaninja, line 1770: types-Flask is dead

hauntsaninja (commented):
Awesome, thanks for adding these! Will take a proper look tomorrow, but noticed one quick thing

hauntsaninja (commented):
At least Auto-Split doesn't work for me:
```
Looking in indexes: https://pypi.org/simple, https://www.riverbankcomputing.com/pypi/simple/
Collecting keyboard
  Cloning https://github.com/Avasam/keyboard.git (to revision fix-563) to /private/var/folders/0y/wt5dx2vj6t16wq97pglrkynm0000gp/T/pip-install-_f3y129x/keyboard_8be85c17f086430d957ec5f247e3481d
  Resolved https://github.com/Avasam/keyboard.git to commit a68455516229d23b19c8fed16cef96eeffa299af
  Preparing metadata (setup.py): started
  Preparing metadata (setup.py): finished with status 'error'

  Running command git clone --filter=blob:none --quiet https://github.com/Avasam/keyboard.git /private/var/folders/0y/wt5dx2vj6t16wq97pglrkynm0000gp/T/pip-install-_f3y129x/keyboard_8be85c17f086430d957ec5f247e3481d
  Running command git checkout -b fix-563 --track origin/fix-563
  Switched to a new branch 'fix-563'
  branch 'fix-563' set up to track 'origin/fix-563'.
  error: subprocess-exited-with-error
  
  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [10 lines of output]
      Traceback (most recent call last):
        File "<string>", line 2, in <module>
        File "<pip-setuptools-caller>", line 34, in <module>
        File "/private/var/folders/0y/wt5dx2vj6t16wq97pglrkynm0000gp/T/pip-install-_f3y129x/keyboard_8be85c17f086430d957ec5f247e3481d/setup.py", line 7, in <module>
          import keyboard
        File "/private/var/folders/0y/wt5dx2vj6t16wq97pglrkynm0000gp/T/pip-install-_f3y129x/keyboard_8be85c17f086430d957ec5f247e3481d/keyboard/__init__.py", line 240, in <module>
          from. import _darwinkeyboard as _os_keyboard
        File "/private/var/folders/0y/wt5dx2vj6t16wq97pglrkynm0000gp/T/pip-install-_f3y129x/keyboard_8be85c17f086430d957ec5f247e3481d/keyboard/_darwinkeyboard.py", line 3, in <module>
          import Quartz
      ModuleNotFoundError: No module named 'Quartz'
      [end of output]
```

If wrangling platform dependencies is tricky, maybe there's some subset of dependencies we can just hardcode that results in good enough type checking.

mypy_primer.py
Avasam: > At least Auto-Split doesn't work for me:
> 
> ```
> Looking in indexes: https://pypi.org/simple, https://www.riverbankcomputing.com/pypi/simple/
> Collecting keyboard
>   Cloning https://github.com/Avasam/keyboard.git (to revision fix-563) to /private/var/folders/0y/wt5dx2vj6t16wq97pglrkynm0000gp/T/pip-install-_f3y129x/keyboard_8be85c17f086430d957ec5f247e3481d
>   Resolved https://github.com/Avasam/keyboard.git to commit a68455516229d23b19c8fed16cef96eeffa299af
>   Preparing metadata (setup.py): started
>   Preparing metadata (setup.py): finished with status 'error'
> 
>   Running command git clone --filter=blob:none --quiet https://github.com/Avasam/keyboard.git /private/var/folders/0y/wt5dx2vj6t16wq97pglrkynm0000gp/T/pip-install-_f3y129x/keyboard_8be85c17f086430d957ec5f247e3481d
>   Running command git checkout -b fix-563 --track origin/fix-563
>   Switched to a new branch 'fix-563'
>   branch 'fix-563' set up to track 'origin/fix-563'.
>   error: subprocess-exited-with-error
>   
>   × python setup.py egg_info did not run successfully.
>   │ exit code: 1
>   ╰─> [10 lines of output]
>       Traceback (most recent call last):
>         File "<string>", line 2, in <module>
>         File "<pip-setuptools-caller>", line 34, in <module>
>         File "/private/var/folders/0y/wt5dx2vj6t16wq97pglrkynm0000gp/T/pip-install-_f3y129x/keyboard_8be85c17f086430d957ec5f247e3481d/setup.py", line 7, in <module>
>           import keyboard
>         File "/private/var/folders/0y/wt5dx2vj6t16wq97pglrkynm0000gp/T/pip-install-_f3y129x/keyboard_8be85c17f086430d957ec5f247e3481d/keyboard/__init__.py", line 240, in <module>
>           from. import _darwinkeyboard as _os_keyboard
>         File "/private/var/folders/0y/wt5dx2vj6t16wq97pglrkynm0000gp/T/pip-install-_f3y129x/keyboard_8be85c17f086430d957ec5f247e3481d/keyboard/_darwinkeyboard.py", line 3, in <module>
>           import Quartz
>       ModuleNotFoundError: No module named 'Quartz'
>       [end of output]
> ```
> 
> If wrangling platform dependencies is tricky, maybe there's some subset of dependencies we can just hardcode that results in good enough type checking.

Looks like you're on macos (which is indeed not supported yet). Quartz module comes from pyobj (Objective-C bindings). Which seems to be a dependency of the keyboard module on mac.
`pyojc ; sys_platform=="darwin"` should do the trick.
Do you mean hardcode here or in AutoSplit?

Avasam:
> At least Auto-Split doesn't work for me:
> [...]
> If wrangling platform dependencies is tricky, maybe there's some subset of dependencies we can just hardcode that results in good enough type checking.


Looks like you're on macos (which is indeed not supported yet). Quartz module comes from pyobj (Objective-C bindings). Which seems to be a dependency of the keyboard module on mac.
`pyojc ; sys_platform=="darwin"` should do the trick.
Do you mean hardcode here or in AutoSplit?

hauntsaninja:
I was referring to hardcoding things in mypy_primer. Your requirements.txt is a little large, e.g. it's got multiple git installs and extra index urls which pip is slow about, and it looks like most of the deps don't have types, so it could make sense to just hardcode things here instead of installing all of that and playing whackamole with the platform specific deps.

mypy_primer.py
Avasam: This and the keyboard issue are fixed now. Can you try again?

mypy_primer.py
```diff
@@ -1763,6 +1775,19 @@ def inner() -> int | None:
         pip_cmd="{pip} install pydantic beartype hydra-core",
         cost=30,
     ),
+    Project(
+        location="https://github.com/Avasam/Auto-Split",
+        mypy_cmd="{mypy} src",
+        pip_cmd="{pip} install -r scripts/requirements.txt types-d3dshot types-keyboard"
```
AlexWaygood, line 1781: I believe this is what @hauntsaninja is suggesting when he says you could "hardcode" the dependencies here -- most of the dependencies you have in your `scripts/requirements.txt` file don't have inline types (so they're irrelevant to type-checking), and will only slow `mypy_primer` down. By just "hardcoding" the necessary dependencies in here, we can speed up mypy_primer and also solve the platform-compatibility issue for macOS :)

```suggestion
        pip_cmd="{pip} install -r numpy certifi packaging types-d3dshot types-keyboard"
```

AlexWaygood: (You can probably do a similar thing for some of the other projects you're adding here as well :)

mypy_primer.py
```diff
@@ -1778,14 +1778,14 @@ def inner() -> int | None:
...
-        " types-toml",
+        pip_cmd="{pip} install certifi ImageHash numpy packaging PyQt6"
+        " types-d3dshot types-keyboard types-Pillow types-psutil types-PyAutoGUI types-pyinstaller"
+        " types-pywin32 types-requests types-toml",
     ),
     Project(
         location="https://github.com/Avasam/speedrun.com_global_scoreboard_webapp",
         mypy_cmd="{mypy} backend",
-        pip_cmd="{pip} install -r scripts/requirements.txt"
+        pip_cmd="{pip} install Flask PyJWT requests-cache sqlalchemy"
```
AlexWaygood, line 1788: SQLAlchemy dosn't have inline types (yet; they're adding them for the next major version bump) -- maybe `types-SQLAlchemy`?

```suggestion
        pip_cmd="{pip} install Flask PyJWT requests-cache types-SQLAlchemy"
```

Avasam: `types-SQLAlchemy` is required by `types-Flask-SQLAlchemy`. I'll just remove that entry.

AlexWaygood (approved):
LGTM!

hauntsaninja:
Thank you! :-)

PR #77, opened 2023-03-02
Add `meson`

eli-schwartz:
Related to #42

See https://github.com/mesonbuild/meson/pull/10994 -- this presumably could have caught https://github.com/python/mypy/issues/12483

(I just merged code into meson that allows passing through the mypy executable and the primer options to our run_mypy.py wrapper, so that this can work. I hope it's not a problem to use our script -- it encodes the names of each file that we have passing mypy, so the run is successful.)

### DIFF
--- a/mypy_primer.py
+++ b/mypy_primer.py
@@ -1597,6 +1597,12 @@ PROJECTS = [
         pip_cmd="{pip} install types-setuptools types-pywin32 types-python-xlib",
         expected_success=True,
     ),
+    Project(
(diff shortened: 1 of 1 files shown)


hauntsaninja (approved):
Thanks, looks like you're using `parse_known_args`, which is the only thing I'd want since you're using a wrapper script

eli-schwartz:
Yeah, I had to specially hook that up for mypy_primer in https://github.com/mesonbuild/meson/pull/11486 :)

PR #79, opened 2023-04-24
(🎁) Show time diff in output

KotlinIsland:
Looks a little like this https://github.com/KotlinIsland/basedmypy/pull/357#issuecomment-1501731245

### DIFF
--- a/mypy_primer.py
+++ b/mypy_primer.py
@@ -545,13 +545,13 @@ class PrimerResult:
         return "\n".join(output_lines)
 
     def header(self) -> str:
-        ret = f"\n{Style.BOLD}{self.project.name}{Style.RESET}\n"
+        ret = f"\n{Style.BOLD}{self.project.name}{Style.RESET} {self.old_result.runtime:.2f}s / {self.new_result.runtime:.2f}s\n"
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
I don't think showing the raw times is a good idea; timing is too finicky and it may vary randomly based on whatever GH Actions is up to.

To do this well, you'd need some real benchmarking setup (like what pyperf does), but that requires many runs and would make mypy-primer slower overall.

JukkaL:
Could we highlight major regression at least? E.g. if the runtime gets 5x slower, it's probably a real regression. This might still be too flaky to be useful, but it seems worth a try.

hauntsaninja:
Yeah, I've been meaning to do this, that's why I added the runtime measurements. I think I fixed a bug and the timing differences aren't too bad, I'll make a version of this that only shows perf differences >10%

hauntsaninja:
Okay, I pushed a version of this in https://github.com/hauntsaninja/mypy_primer/commit/bedddb75e10fe0f2cbd01e5d0b4f3b6098d89e0c

Screenshot 2023-04-25 at 3 28 16 PM

Thanks for bumping this!

PR #83, opened 2023-06-06
add rudimentary pyright support

hauntsaninja:
This pull request follows up on a bunch of refactoring from earlier today.

This is enough to make basic commands like `mypy_primer --type-checker pyright --old a08f716f9 --debug -k hauntsaninja/pyp` do something.

There's still basic stuff left to be done. For instance, get pyright to use the project virtual environment and custom typeshed dir and alter project list logic for pyright's needs.

### DIFF
--- a/mypy_primer/globals.py
+++ b/mypy_primer/globals.py
@@ -9,10 +9,11 @@ import mypy_primer.utils
 
 @dataclass
 class _Args:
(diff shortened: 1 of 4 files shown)


hauntsaninja:
cc @erictraut I've also given you access to this repository

PR #84, opened 2023-06-08
Added more pyright functionality including a filter for projects that…

erictraut:
… are known to work with pyright and projects that are expected to type check without errors in pyright.


erictraut:
I need to fix some merge conflicts. Will resubmit PR when those are fixed.

PR #89, opened 2023-06-14
Autosplit: don't install winsdk

AlexWaygood:
Fixes https://github.com/python/mypy/pull/15429#issuecomment-1590951690

Cc. @Avasam


Avasam:
Oops, thanks!
That"s why it wasn't there >.<

PR #90, opened 2023-06-27
update cwltool/schema-salad setup to be more resiliant

mr-c:
`expected_mypy_success=True` as well


mr-c:
Tested locally with `mypy_primer --debug --type-checker mypy -k common-workflow-language`

mr-c:
This is ready for review and merging, thanks!

A5rocks:
I'm not sure `grep` here is allowed because mypy-primer *should* (though it's kinda bad at it) be able to run on Windows.

hauntsaninja (commented):
Thanks, would you mind marking it using the nice `supported_platforms` annotation A5rocks added in https://github.com/hauntsaninja/mypy_primer/pull/110/files

mr-c:
@hauntsaninja Done, thank you!

hauntsaninja (approved):

PR #91, opened 2023-06-29
doc: clarify the name origin

not-my-profile:
I was confused how "black" inspired "primer" ... I think it makes sense to clarify that :)


hauntsaninja:
Thanks, that's right. The README's gotten moved around though, so closing this :-)

PR #94, opened 2023-07-27
Make the projects be their own `repr`!

A5rocks:
~~There's still a couple of things that are in the way (specifically comments and 1 specific version-specific project). Not sure how to handle that. Also `black` doesn't split string lines which means annoying flake8 :(~~ fixed.

Once the above is fixed, this means the project list can be modified at runtime e.g. to update dependencies or to cache some information (like runtime cost).

### DIFF
--- a/mypy_primer/projects.py
+++ b/mypy_primer/projects.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import difflib
+import subprocess
 import sys
 
 from mypy_primer.model import Project
@@ -33,11 +34,12 @@ def update_projects(projects: list[Project], check: bool = False) -> None:
                 keep = True
 
     if check:
-        import black
-
-        code = black.format_file_contents(
-            "\n".join(result), fast=False, mode=black.mode.Mode(preview=True, line_length=100)
+        code_proc = subprocess.run(
+            ["black", "-c", "\n".join(result)], capture_output=True, text=True
         )
+        code_proc.check_returncode
+        code = code_proc.stdout
+
         with open(__file__) as f:
             in_file = f.read()
             if in_file != code:
--- a/mypy_primer/projects.py
+++ b/mypy_primer/projects.py
@@ -1,7 +1,5 @@
 from __future__ import annotations
 
-import sys
-
 from mypy_primer.model import Project
 
 # projects to re-enable:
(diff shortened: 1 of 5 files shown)


mypy_primer/projects.py
```diff
@@ -1,9 +1,58 @@
...
+                result.append(line)
+            if line == "    ]":
+                keep = True
+
+    if check:
+        import black
+
+        code = black.format_file_contents(
+            "\n".join(result), fast=False, mode=black.mode.Mode(preview=True, line_length=100)
+        )
```
A5rocks, line 40: Is this really how I'm supposed to use this?

(TBF it might be a nice idea to black-format mypy unit tests?)

JelleZijlstra: You're not supposed to do any of this: psf/black#779 :) However, this particular call is likely to keep working.

I am not sure I understand why we need to import black here, though, or even what exactly this code is trying to do; any reason not to invoke `black --diff` in a subprocess?

A5rocks: Specifically this code is trying to avoid writing out to the filesystem.

JelleZijlstra:
```
b$ black --diff -c 'x=3'
--- STDIN       2023-07-27 01:59:48.499661+00:00
+++ STDOUT      2023-07-27 01:59:48.503988+00:00
@@ -1 +1 @@
-x=3
\ No newline at end of file
+x = 3
```

A5rocks: Oh, I didn't realize you could do that! I'll look at that

A5rocks: Err, actually, that doesn't quite do what I want. Basically, what checking does is:
1. replaces `projects = [...]` with the repr'd version of itself
2. runs `black` over that new (atm in-memory only) file
3. compares that new output to the old file contents

Now that I think of it, I can just use `black -c '...'` instead.......

A5rocks:
I really shouldn't be writing these PRs when tired :(

Thank you for the review @Akuli. Slightly unrelated question, cause I don't know where else to ask: is there a reason you archived https://github.com/Akuli/issue-downloader? Do you still think something like that is plausible?

Akuli:
TIL that you can't create issues on archived repos :)

It was some combination of these, don't remember what exactly:
- GitHub API limits: can only download n issues per hour. Worse when you do it locally and that's why I did it in github actions, I think.
- GitHub Actions limitations: I think making the repo public was enough to get around this
- mypy can only take n files as command-line arguments at once. Can't do `mypy huge_folder/*.py` and instead need to run mypy on them one by one, which was slow.

It might be possible to get that to work, feel free to try if you have time :)

hauntsaninja (approved):
Thank you!

PR #97, opened 2023-08-30
Add three mypy-checked projects from CPython

AlexWaygood:
Argument Clinic, the cases generator, and the peg generator are now all checked with mypy in CI. All have very strict settings enabled in their mypy configs (including the new `possibly-undefined` error code for the cases generator). The cases generator and Argument Clinic also make heavy use of pattern-matching; and the peg generator makes some use of relatively obscure parts of `setuptools` (which is useful for typeshed).

### DIFF
--- a/mypy_primer/git_utils.py
+++ b/mypy_primer/git_utils.py
@@ -12,8 +12,8 @@ RevisionLike = Union[str, None, Callable[[Path], Awaitable[str]]]
 
 
 async def clone(repo_url: str, repo_dir: Path, cwd: Path, shallow: bool = False) -> None:
-    if os.path.exists(repo_url):
-        repo_url = os.path.abspath(repo_url)
+    if os.path.exists(repo_dir):
+        repo_url = os.path.abspath(repo_dir)
     cmd = ["git", "clone", "--recurse-submodules", repo_url, str(repo_dir)]
     if shallow:
         cmd += ["--depth", "1"]
--- a/mypy_primer/git_utils.py
+++ b/mypy_primer/git_utils.py
@@ -11,10 +11,11 @@ from mypy_primer.utils import run
 RevisionLike = Union[str, None, Callable[[Path], Awaitable[str]]]
 
 
-async def clone(repo_url: str, repo_dir: Path, cwd: Path, shallow: bool = False) -> None:
-    if os.path.exists(repo_url):
-        repo_url = os.path.abspath(repo_url)
-    cmd = ["git", "clone", "--recurse-submodules", repo_url, str(repo_dir)]
+# repo_source could be a URL *or* a local path
+async def clone(repo_source: str, *, repo_dir: Path, cwd: Path, shallow: bool = False) -> None:
+    if os.path.exists(repo_source):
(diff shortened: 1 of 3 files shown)


AlexWaygood:
The changes in `model.py` were required to avoid tripping this assertion here, since this adds three projects from the same repo:

https://github.com/hauntsaninja/mypy_primer/blob/cd38346c2e231d4ec015d4b57416426cff692ab2/mypy_primer/projects.py#L863

(Very happy to go with a different fix if you don't like this one!)

AlexWaygood:
Note: although these projects live in the CPython repo, they're guaranteed to not use any syntax newer than Python 3.10. That's because these projects are used to build CPython itself, and CPython 3.13 can't rely on features new-in-py312 in its build scripts, or it would run into bootstrapping issues. As such, all three projects have `python_version = 3.10` in their mypy config files.

hauntsaninja (changes requested):
This seems reasonable. The reason the uniqueness check exists is to ensure that things are all getting checked out to different locations. I think it might be worth checking these three out to different locations as well (it might work as is, but maybe concurrent git clones would go badly and it would definitely break if we e.g. pinned one of these to a revision or something like that). Just need to thread the name override through to `ensure_repo_at_revision` as well

mypy_primer/git_utils.py
```diff
@@ -11,10 +10,10 @@
 RevisionLike = Union[str, None, Callable[[Path], Awaitable[str]]]
 
 
-async def clone(repo_url: str, cwd: Path, shallow: bool = False) -> None:
-    if os.path.exists(repo_url):
-        repo_url = os.path.abspath(repo_url)
-    cmd = ["git", "clone", "--recurse-submodules", repo_url]
+async def clone(repo_url: str, repo_dir: Path, cwd: Path, shallow: bool = False) -> None:
+    if repo_dir.exists():
+        repo_dir = repo_dir.absolute()
```
AlexWaygood, line 15: Not sure I fully understand what's going on here; this might be incorrect

hauntsaninja: Yeah, this isn't right. The tricky thing here is that `repo_url` doesn't *have* to be a URL, it can also be a local path. This is useful for e.g. running against a change in my local typeshed checkout. I definitely should have added a comment though!

AlexWaygood: Aha, I see! Would you be up for renaming the parameter something like `repo_location`?

hauntsaninja (approved):
Looks good! One nit is maybe we can make some clone args kw-only (I like using kw-only args when things have the same type like cwd and repo_dir)

Hmm repo_location sounds a little big ambiguous. Maybe repo_source?

AlexWaygood:
Okay cool, I think I've fixed the final nits!

hauntsaninja:
Thank you!

hauntsaninja:
Temporarily reverted in https://github.com/hauntsaninja/mypy_primer/commit/3fe90cf5c76de767ac4f1e46c9893a0dde0376cd in case it's causing issues

PR #99, opened 2023-09-09
Make primer work on Windows again

AlexWaygood:
I set out to figure out why the changes I made in #97 caused primer to hang in typeshed's CI, but quickly realised that I couldn't run primer locally at all any more! This is because `shlex.quote()` does the wrong thing on Windows when you pass it a path. All you actually need to do on Windows to quote a _path_ is to do some basic validation and then surround it with `""`, so there's a pretty simple workaround.

### DIFF
--- a/mypy_primer/model.py
+++ b/mypy_primer/model.py
@@ -96,7 +96,7 @@ class Project:
             venv.create(self.venv_dir, with_pip=True, clear=True)
             try:
                 await run(
-                    self.pip_cmd.format(pip=str(self.venv_dir / BIN_DIR / "pip")),
+                    self.pip_cmd.format(pip=quote_path(self.venv_dir / BIN_DIR / "pip")),
                     shell=True,
                     cwd=repo_dir,
                     output=True,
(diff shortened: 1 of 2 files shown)


hauntsaninja (commented):
Huh, TIL. Is the issue that shlex.quote prefers single quotes instead of double quotes?

AlexWaygood:
> Is the issue that shlex.quote prefers single quotes instead of double quotes?

I think that's the main issue, yeah, though I don't really know much more than what [the warning in the docs](https://docs.python.org/3/library/shlex.html#shlex.quote) tells me.

This solution is derived from [Eryk Sun's comments on a Stack Overflow question](https://stackoverflow.com/questions/61691532/how-to-quote-paths-on-windows-similar-to-shlex-quote#comment109123375_61691532) that suggest that this should always do the right thing if all you need to do is to quote a Windows path, and it was the quoting of paths that was causing me grief locally.

An alternative solution would be to use `subprocess.list2cmdline`, which I think tries harder to capture all the semantics of the Windows command line. I didn't go for that because:

1. Not sure we need it here 
2. It's undocumented
3. There was some scepticism recently when people suggested documenting it: https://discuss.python.org/t/why-is-subprocess-list2cmdline-not-documented/25044/1

hauntsaninja (approved):
Thanks for the links!

PR #101, opened 2023-09-15
Once again, add 3 mypy-checked projects from CPython

AlexWaygood:
This is a redo of #97, which was reverted because it seemed to be causing CI in typeshed and mypy to hang. I'm still not sure exactly what was causing primer to hang in CI, but I can't reproduce it locally. I also can't reproduce it in CI anymore: I tested out this patch using my typeshed fork here, and it seems to be working fine:

- https://github.com/AlexWaygood/typeshed/pull/20

(Ignore all the failing CI on ^that demo PR -- I cancelled most of the non-primer jobs as there didn't seem to be any point wasting the CI minutes when I was only interested in mypy_primer.)

### DIFF
--- a/mypy_primer/git_utils.py
+++ b/mypy_primer/git_utils.py
@@ -11,10 +11,11 @@ from mypy_primer.utils import run
 RevisionLike = Union[str, None, Callable[[Path], Awaitable[str]]]
 
 
-async def clone(repo_url: str, cwd: Path, shallow: bool = False) -> None:
-    if os.path.exists(repo_url):
-        repo_url = os.path.abspath(repo_url)
-    cmd = ["git", "clone", "--recurse-submodules", repo_url]
+# repo_source could be a URL *or* a local path
+async def clone(repo_source: str, *, repo_dir: Path, cwd: Path, shallow: bool = False) -> None:
+    if os.path.exists(repo_source):
+        repo_source = os.path.abspath(repo_source)
+    cmd = ["git", "clone", "--recurse-submodules", repo_source, str(repo_dir)]
     if shallow:
         cmd += ["--depth", "1"]
     await run(cmd, cwd=cwd)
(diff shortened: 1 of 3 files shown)


hauntsaninja:
Hmm, still a little worrying. Maybe we should try harder to repro the hang, e.g. run with fewer shards and fewer non-CPython projects

AlexWaygood:
FWIW, the command I ran locally to test was

```
mypy_primer --custom-typeshed-repo ../typeshed --new-typeshed a4f08bd --old-typeshed 4ae6d38 --new v1.5.1 --old v1.5.1 -o full --debug -k CPython
```

AlexWaygood:
I think it's possible the issue may have been something to do with the fact that these project names have whitespace in them. It's possible that that could have been fixed by e.g. https://github.com/hauntsaninja/mypy_primer/commit/d3a8fb26c0bbcbab07e58d90e0aa62142c27bd42#diff-98ad6c102b483c2dd09ca46d76ecab89be0699a334f207fd9f8790b16f2c0266R98-R99

AlexWaygood:
> I think the issue _may_ have been something to do with the fact that these project names have whitespace in them. It's _possible_ that that could have been fixed by e.g. [d3a8fb2#diff-98ad6c102b483c2dd09ca46d76ecab89be0699a334f207fd9f8790b16f2c0266R98-R99](https://github.com/hauntsaninja/mypy_primer/commit/d3a8fb26c0bbcbab07e58d90e0aa62142c27bd42#diff-98ad6c102b483c2dd09ca46d76ecab89be0699a334f207fd9f8790b16f2c0266R98-R99)
> 
> Let me see...

I think this is the correct diagnosis.

In https://github.com/AlexWaygood/typeshed/pull/20, I'm now reproducing the hang by switching the branch of `AlexWaygood/mypy_primer` I'm using in CI to https://github.com/AlexWaygood/mypy_primer/tree/cpython-again-test. That branch is identical to this one except for https://github.com/AlexWaygood/mypy_primer/commit/2d1fe5e59d3be117cee2f0d376c8005f85ea670a (which is a revert of https://github.com/hauntsaninja/mypy_primer/commit/d3a8fb26c0bbcbab07e58d90e0aa62142c27bd42#diff-98ad6c102b483c2dd09ca46d76ecab89be0699a334f207fd9f8790b16f2c0266R98-R99). It appears that the primer shard that's doing the CPython projects has started hanging almost immediately after printing

```
/tmp/mypy_primer/projects/_CPython (peg_generator)_venv/bin/pip install types-setuptools types-psutil	 in /tmp/mypy_primer/projects/CPython (peg_generator)
```

hauntsaninja (approved):
Awesome, thanks so much for figuring that out, and time traveling to fix it!

AlexWaygood:
Pleasure! :D

PR #103, opened 2023-10-14
Include annotation-tests in hydra-zen

rsokl:
hydra-zen uses tests/annotations/ to test pyright against a wide variety of common usage patterns, which have frequently caught regressions in new pyright releases. This PR includes adds these files to the pyright scan of hydra-zen.

See [discussion with Eric](https://github.com/microsoft/pyright/issues/6071#issuecomment-1745229771)

### DIFF
--- a/mypy_primer/projects.py
+++ b/mypy_primer/projects.py
@@ -827,7 +827,7 @@ def get_projects() -> list[Project]:
         ),
         Project(
             location="https://github.com/mit-ll-responsible-ai/hydra-zen",
-            mypy_cmd="{mypy} src  tests/annotations/mypy_checks.py",
+            mypy_cmd="{mypy} src tests/annotations/mypy_checks.py",
             pyright_cmd="{pyright} tests/annotations src",
             pip_cmd="{pip} install pydantic beartype hydra-core",
             mypy_cost=30,
--- a/mypy_primer/projects.py
+++ b/mypy_primer/projects.py
@@ -828,7 +828,7 @@ def get_projects() -> list[Project]:
         Project(
             location="https://github.com/mit-ll-responsible-ai/hydra-zen",
             mypy_cmd="{mypy} src",
-            pyright_cmd="{pyright}",
+            pyright_cmd="{pyright} tests/annotations/ src/",
             pip_cmd="{pip} install pydantic beartype hydra-core",
             mypy_cost=30,
         ),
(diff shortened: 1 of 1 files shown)


mypy_primer/projects.py
```diff
@@ -828,7 +828,7 @@ def get_projects() -> list[Project]:
         Project(
             location="https://github.com/mit-ll-responsible-ai/hydra-zen",
             mypy_cmd="{mypy} src",
```
erictraut, line 830: Would you like to include `tests/annotations` for mypy as well for symmetry and completeness?

mypy_primer/projects.py
```diff
@@ -828,7 +828,7 @@ def get_projects() -> list[Project]:
...
-            pyright_cmd="{pyright}",
+            pyright_cmd="{pyright} tests/annotations/ src/",
```
erictraut, line 831: Minor nit: The trailing slashes shouldn't be needed here. Maybe remove for consistency.

erictraut (commented):
Thanks for your contribution, @rsokl!

mypy_primer/projects.py
rsokl: I generally do not run mypy against that file in my CI because there are too many features/patterns that mypy does not support (at least historically). I only have the bandwidth to ensure high-fidelity support for pyright. 

That being said, I do have basic patterns that I run mypy against, which I will include.

rsokl, line 830:
```suggestion
            mypy_cmd="{mypy} src  tests/annotations/mypy_checks.py",
```

mypy_primer/projects.py
rsokl, line 831:
```suggestion
            pyright_cmd="{pyright} tests/annotations src",
```

mypy_primer/projects.py
```diff
@@ -827,8 +827,8 @@ def get_projects() -> list[Project]:
         ),
         Project(
             location="https://github.com/mit-ll-responsible-ai/hydra-zen",
-            mypy_cmd="{mypy} src",
-            pyright_cmd="{pyright}",
+            mypy_cmd="{mypy} src  tests/annotations/mypy_checks.py",
```
rsokl, line 830:
```suggestion
            mypy_cmd="{mypy} src tests/annotations/mypy_checks.py",
```

erictraut (approved):

hauntsaninja:
Thank you!

PR #109, opened 2023-11-26
Add more dependencies for meson

eli-schwartz:
Our configuration sets ignore_missing_imports so it isn't immediately apparent these would be used...


eli-schwartz:
See also https://github.com/mesonbuild/meson/pull/12551 (fixes our mypy.ini since it seems like setting that was always wrong; will result in failing to typecheck without installing these deps).

hauntsaninja:
Thanks, I fixed the conflict and merged to master

PR #110, opened 2023-11-26
Fix windows support for `mypy_primer --coverage`

A5rocks:
This makes `mypy_primer --coverage` run without errors... though the output is wrong:

```
Checking 131 projects...
Containing 0 files...
Totalling to 0 lines...
```

Good enough.

### DIFF
--- a/mypy_primer/projects.py
+++ b/mypy_primer/projects.py
@@ -112,7 +112,7 @@ def get_projects() -> list[Project]:
             mypy_cmd="{mypy} aiohttp",
             pip_cmd="AIOHTTP_NO_EXTENSIONS=1 {pip} install -e . pytest",
             expected_mypy_success=True,
-            only_platform="linux",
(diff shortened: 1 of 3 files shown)


mypy_primer/projects.py
```diff
@@ -144,6 +145,7 @@ def get_projects() -> list[Project]:
             mypy_cmd="{mypy} boostedblob",
             pip_cmd="{pip} install aiohttp uvloop pycryptodome",
             expected_mypy_success=True,
+            only_platform="linux",
```
A5rocks, line 148: Would this be fine?:

```suggestion
            pip_cmd="{pip} install aiohttp pycryptodome",
            expected_mypy_success=True,
```

hauntsaninja (approved):
Thanks, I pushed a change to allow multiple platforms / allow these to run on Mac. If it looks good to you, I'll merge

PR #114, opened 2024-01-31
install pluggy in CI when running mypy on pytest

AlexWaygood:
https://github.com/python/typeshed/pull/11351#issuecomment-1919424187


hauntsaninja (approved):
Yeah probably new Black, I'll fix...

PR #119, opened 2024-03-16
Added setuptools & pkg_resources

Avasam:
`setuptools` & `pkg_resources` are important users of the stdlib, and are checking against mypy as of https://github.com/pypa/setuptools/pull/4257
We can also use `setuptools`' own tests to validate `types-setuptools` changes


hauntsaninja (approved):
Thanks!

PR #121, opened 2024-04-19
Add support for type checking with mypy plugins

### LINKED ISSUE #8: How to add dependencies with plugins?

### LINKED ISSUE #102: do a plugin hack
this is basically a dupe of https://github.com/hauntsaninja/mypy_primer/issues/8 , but we're currently not type checking home-assistant/core and this is usually quite a useful repo to validate against

flaeppe:
~I'm not totally sure if this approach always works or if there's stuff that I'm missing, but it seems to work. I'll try to explain myself here.~

~The change utilises `pip`'s `--target`([docs](https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-t)) argument to point out where to install a project's additional dependencies, passed in via `pip_cmd`. This replaces the virtual env creation for each project, since that shouldn't be necessary (I think?).~

~In addition to what's mentioned above~ we start setting the `PYTHONPATH` env variable([docs](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONPATH)), to point out an additional search path for module files, namely the site-packages directory ~with the additional dependencies~ inside the project's virtualenv.

That should allow discovery of project dependencies while keeping the installed mypy versions independent and reusable.

- Closes #8 
- Closes #102 

<details>
  <summary>Outdated</summary>

---

All this _isn't_ included, but I was thinking that it _should_ be possible to use `pip install --target` for both mypy installations as well, instead of a virtual env. That leads to having 1 python/pip executable, 1 directory for `new_mypy`, 1 directory for `old_mypy` and then all project dependency installations. Although I'm unsure what `--python-executable=` does and if it could be affected. I'm also not sure how to handle compiled mypy.

Essentially what'll happen with the above changes is that we'll have a directory structure like e.g.

```bash
repo-dir/
├── new_mypy
│   └── bin
│       └── mypy
├── old_mypy
│   └── bin
│       └── mypy
├── projects
│   ├── A
│   └── B
└── python
    └── bin
        └── python
```

We run mypy with something corresponding to

```
PYTHONPATH="repo-dir/new_mypy:repo-dir/projects/A" repo-dir/new_mypy/bin/mypy --python-executable=repo-dir/python/bin/python ...
PYTHONPATH="repo-dir/old_mypy:repo-dir/projects/A" repo-dir/old_mypy/bin/mypy --python-executable=repo-dir/python/bin/python ...
```
</details>

---

Here's some example debug output:

```
...
/tmp/mypy_primer/projects/_django-choicefield_venv/bin/pip install . django-stubs pytest         in /tmp/mypy_primer/projects/django-choicefield
PYTHONPATH=/tmp/mypy_primer/new_mypy/venv/lib/python3.12/site-packages:/tmp/mypy_primer/projects/_django-choicefield_venv/lib/python3.12/site-packages
/tmp/mypy_primer/new_mypy/venv/bin/mypy --python-executable=/tmp/mypy_primer/projects/_django-choicefield_venv/bin/python --warn-unused-ignores --warn-redundant-casts --no-incremental --cache-dir=/dev/null --show-traceback --soft-error-limit=-1     in /tmp/mypy_primer/projects/django-choicefield
PYTHONPATH=/tmp/mypy_primer/old_mypy/venv/lib/python3.12/site-packages:/tmp/mypy_primer/projects/_django-choicefield_venv/lib/python3.12/site-packages
/tmp/mypy_primer/old_mypy/venv/bin/mypy --python-executable=/tmp/mypy_primer/projects/_django-choicefield_venv/bin/python --warn-unused-ignores --warn-redundant-casts --no-incremental --cache-dir=/dev/null --show-traceback --soft-error-limit=-1     in /tmp/mypy_primer/projects/django-choicefield
/tmp/mypy_primer/new_mypy/venv/bin/mypy on django-choicefield took 4.96s
/tmp/mypy_primer/old_mypy/venv/bin/mypy on django-choicefield took 4.98s
...
```

### DIFF
--- a/mypy_primer/model.py
+++ b/mypy_primer/model.py
@@ -16,6 +16,7 @@ from typing import Sequence
 
 from mypy_primer.git_utils import ensure_repo_at_revision
 from mypy_primer.globals import ctx
+from mypy_primer.type_checker import Checker
 from mypy_primer.utils import BIN_DIR, Style, debug_print, quote_path, run
 
 extra_dataclass_args = {"kw_only": True} if sys.version_info >= (3, 10) else {}
@@ -138,7 +139,7 @@ class Project:
         )
         return mypy_cmd
 
-    async def run_mypy(self, mypy: str | Path, typeshed_dir: Path | None) -> TypeCheckResult:
+    async def run_mypy(self, mypy: Checker, typeshed_dir: Path | None) -> TypeCheckResult:
         additional_flags = ctx.get().additional_flags.copy()
         env = os.environ.copy()
         env["MYPY_FORCE_COLOR"] = "1"
@@ -153,9 +154,11 @@ class Project:
         env["MYPYPATH"] = os.pathsep.join(mypy_path)
         pythonpath = env.get("PYTHONPATH", "").split(os.pathsep)
         pythonpath.insert(0, str(self.venv_dir))
+        if mypy.site_packages is not None:
+            pythonpath.insert(0, str(mypy.site_packages))
         env["PYTHONPATH"] = os.pathsep.join(pythonpath)
 
-        mypy_cmd = self.get_mypy_cmd(mypy, additional_flags)
+        mypy_cmd = self.get_mypy_cmd(mypy.path, additional_flags)
         if ctx.get().debug:
             debug_print(f'{Style.BLUE}PYTHONPATH={env["PYTHONPATH"]}{Style.RESET}')
         proc, runtime = await run(
@@ -167,7 +170,7 @@ class Project:
             env=env,
         )
         if ctx.get().debug:
-            debug_print(f"{Style.BLUE}{mypy} on {self.name} took {runtime:.2f}s{Style.RESET}")
+            debug_print(f"{Style.BLUE}{mypy.path} on {self.name} took {runtime:.2f}s{Style.RESET}")
 
         output = proc.stderr + proc.stdout
 
@@ -206,11 +209,11 @@ class Project:
         pyright_cmd = pyright_cmd.format(pyright=pyright)
         return pyright_cmd
 
-    async def run_pyright(self, pyright: str | Path, typeshed_dir: Path | None) -> TypeCheckResult:
+    async def run_pyright(self, pyright: Checker, typeshed_dir: Path | None) -> TypeCheckResult:
         additional_flags: list[str] = []
         if typeshed_dir is not None:
             additional_flags.append(f"--typeshedpath {quote_path(typeshed_dir)}")
-        pyright_cmd = self.get_pyright_cmd(pyright, additional_flags)
+        pyright_cmd = self.get_pyright_cmd(pyright.path, additional_flags)
         if self.pip_cmd:
             activate = (
                 f"source {shlex.quote(str(self.venv_dir / BIN_DIR / 'activate'))}"
@@ -226,7 +229,7 @@ class Project:
             cwd=ctx.get().projects_dir / self.name,
         )
         if ctx.get().debug:
-            debug_print(f"{Style.BLUE}{pyright} on {self.name} took {runtime:.2f}s{Style.RESET}")
+            debug_print(f"{Style.BLUE}{pyright.path} on {self.name} took {runtime:.2f}s{Style.RESET}")
 
         output = proc.stderr + proc.stdout
         return TypeCheckResult(
@@ -234,7 +237,7 @@ class Project:
         )
 
     async def run_typechecker(
-        self, type_checker: str | Path, typeshed_dir: Path | None
+        self, type_checker: Checker, typeshed_dir: Path | None
     ) -> TypeCheckResult:
         if ctx.get().type_checker == "mypy":
             return await self.run_mypy(type_checker, typeshed_dir)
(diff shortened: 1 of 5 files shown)


mypy_primer/model.py
```diff
@@ -153,8 +152,13 @@ async def run_mypy(self, mypy: str | Path, typeshed_dir: Path | None) -> TypeChe
         if "MYPYPATH" in env:
             mypy_path = env["MYPYPATH"].split(os.pathsep) + mypy_path
         env["MYPYPATH"] = os.pathsep.join(mypy_path)
+        pythonpath = env.get("PYTHONPATH", "").split(os.pathsep)
+        pythonpath.insert(0, str(self.venv_dir))
+        env["PYTHONPATH"] = os.pathsep.join(pythonpath)
```
flaeppe, line 157: Forgot to prepend `repo-dir/{new|old}_mypy/venv/lib/python3.XX/site-packages` here, slightly fiddly to get hold of the path though.

Need it in case `mypy` would be installed by any additional deps we've installed, we want precedence on old/new mypy

mypy_primer/model.py
```diff
@@ -127,7 +125,7 @@ def get_mypy_cmd(self, mypy: str | Path, additional_flags: Sequence[str] = ()) -
         mypy_cmd = mypy_cmd.format(mypy=mypy)
 
         if self.pip_cmd:
-            python_exe = self.venv_dir / BIN_DIR / "python"
+            python_exe = ctx.get().venv_dir / BIN_DIR / "python"
```
flaeppe, line 128: Hm, perhaps we shouldn't change the `--python-executable=` binary in this PR.

I'm not at all sure what could happen if we do/don't but it seems to work as intended without change, so we could keep it unchanged

flaeppe: Oh, nvm, I was confusing this with thinking it was set to the python binary within the old/new mypy directories. `self.venv_dir / BIN_DIR / "python"` doesn't exist any longer since we don't have a virtual env for each project.

mypy_primer/model.py
flaeppe: Fixed in 04e79e8dfb29d786b115b175388b9bf1811307fc

mypy_primer/model.py
```diff
@@ -106,14 +106,13 @@ async def setup(self) -> None:
         assert repo_dir == ctx.get().projects_dir / self.name
         if self.pip_cmd:
             assert "{pip}" in self.pip_cmd
-            venv.create(self.venv_dir, with_pip=True, clear=True)
+            pip, install, tail = self.pip_cmd.partition(" install ")
+            cmd = (pip + install + "--target {target} " + tail).format(
+                pip=quote_path(ctx.get().pip_exe),
+                target=quote_path(self.venv_dir),
+            )
```
flaeppe, line 113: Hm since the venv was previously created with `clear=True` it might be that we want to install with either `--force-reinstall` or `--ignore-installed`

hauntsaninja (changes requested):
Thanks for the PR!

I don't quite understand what the new global env is for. Other than that, I think this PR mostly does the right thing: we should continue to setup a) a venv for mypy itself, b) project specific venvs. And then use `PYTHONPATH` (or other) shenanigans so that mypy can import from the project specific venv. We should also add an assert that some version of mypy itself isn't in the project specific venv, otherwise things will get confusing.

flaeppe:
Thanks for the review, I'm gonna try to move back to using project specific venvs, see if I can get it to work.

The global env was added to have a python/pip binary to reach to when I removed the project specific venvs.

I'll see if I can get a reasonable assert in there too, perhaps just something naive like looking for `bin/mypy` in a project specific venv is enough.

flaeppe:
I've updated to align with your suggestions. Let me know if there's anything else I've missed or should update.

Though I can't say I follow why each project needs its own environment. I was thinking it would be preferable to have a virtualenv for each mypy installation and then do a "distributed" install for each project (`pip install --target`) and via `PYTHONPATH=` "mount" in the relevant dependencies for each project run. But I might definitely miss something.

hauntsaninja:
Thanks again for the PR, sorry that I lost track of it!
This developed a few merge conflicts, I fixed in 83721d5. I also altered the approach slightly, I use a `.pth` file to *append* to sys.path, so that project venv will never cause conflicts with mypy and its dependencies.

PR #122, opened 2024-06-11
Remove SinbadCogs

srittau:
The project has been archived upstream. Cc @mikeshardmind


mikeshardmind:
Types aren't actively maintained on it anymore, so it's definitely worth removal. Some of what were the more complex parts of that repo still exist in some form in either https://github.com/mikeshardmind/async-utils or https://github.com/mikeshardmind/discord-scheduler if you're interested in keeping any specific bits of the async logic that was well typed in primer runs, but I'm not significantly concerned that anything in these is likely to break in a way other things in the primer runs won't catch with the number of projects now included.

PR #124, opened 2024-07-23
Add `django-stubs` to the list of projects

sobolevn:
Let's try it!

### DIFF
--- a/mypy_primer/projects.py
+++ b/mypy_primer/projects.py
@@ -17,6 +17,7 @@ from mypy_primer.model import Project
 # - https://github.com/NeilGirdhar/efax
 # - https://github.com/google/duet
 # - https://github.com/RobertCraigie/prisma-client-py
+# - https://github.com/typeddjango/django-stubs
 
 
 def update_projects(projects: list[Project], check: bool = False) -> None:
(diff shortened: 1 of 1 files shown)


hauntsaninja (changes requested):
Hm, I think you'll need some sort of install command to install the plugin you want to use while checking django-stubs (possible just `install_cmd="{install} .",`)

sobolevn:
@hauntsaninja how can I test that the whole intergration worked?

sobolevn:
Thanks for the quick review and merge! ❤️

hauntsaninja:
Yes, that'll work. I also just ran it locally before merging using `mypy_primer -k django-stub --debug --old v1.10.0`

AlexWaygood:
Some strange resolution failure here when trying to install django-stubs: https://github.com/python/typeshed/actions/runs/10112015097/job/27965167268?pr=12433.

Hopefully just a fluke of some kind, I'll try rerunning

sobolevn:
@flaeppe looks like our release is partially broken :(

flaeppe:
I think we just need to make a release. We only prepared for one.

sobolevn:
oh, I see then why it fails: we bumped the dependency in `git` and not release `ext@5.0.3`

flaeppe:
But something seems quite off if we're installing `.` on `master` and run in to troubles. When installing in our CI, with `pip`, we set the env variable `SETUPTOOLS_ENABLE_FEATURES=legacy-editable` not sure if that affects anything

andersk:
We need to install both `.` and `./ext`, not just `.`.

- #125

PR #130, opened 2024-12-24
Create pylint.yml

alfapppro:


JelleZijlstra:
I don't think we need pylint on this repo.

PR #131, opened 2025-01-06
Add setuptools deps for pkg_resources

Avasam:
Libraries that typecheck their usage of `pkg_resources`.
This should resolve the primer hits from removing `pkg_resources` in typeshed: https://github.com/python/typeshed/pull/13369#issuecomment-2571843734


mypy_primer/projects.py
```diff
@@ -560,6 +560,7 @@ def get_projects() -> list[Project]:
             location="https://github.com/laowantong/paroxython",
             mypy_cmd="{mypy} paroxython",
             pyright_cmd="{pyright} paroxython",
+            deps=["setuptools>=71.1"],
```
Avasam, line 563: Apparently this doesn't hit anymore: https://github.com/python/typeshed/pull/13369#issuecomment-2655793138

mypy_primer/projects.py
```diff
@@ -1415,7 +1415,7 @@ def get_projects() -> list[Project]:
             mypy_cmd="{mypy} {paths}",
             pyright_cmd="{pyright} {paths}",
             paths=["pwndbg"],
-            deps=["types-gdb"],
+            deps=["setuptools>=71.1", "types-gdb"],
```
Avasam, line 1418: `pkg_resources` import removed in https://github.com/pwndbg/pwndbg/pull/1983
```suggestion
            deps=["types-gdb"],
```

PR #132, opened 2025-02-14
Add `setuptools` to `deps` of `scipy` and `pwndbg`

Avasam:
According to the primer hits in https://github.com/python/typeshed/pull/13369#issuecomment-2655793138 , these libs typecheck files that import `pkg_resources`


Avasam:
Woops, I duplicated my own PR...
https://github.com/hauntsaninja/mypy_primer/pull/131

PR #135, opened 2025-03-12
Project: use list of paths instead of `pyright_cmd`

sharkdp:
Thank you for creating `mypy_primer`! As of recently, we use it to run test [Red Knot](https://github.com/astral-sh/ruff/blob/main/crates/red_knot/README.md) against (a small subset of) ecosystem projects. So far, we use a [fork](https://github.com/astral-sh/mypy_primer/tree/add-red-knot-support) of `mypy_primer`, but I would like to clean up our implementation, add some improvements, and potentially upstream some of the changes — if you are interested.

As a first step, I would like to propose a change to the `Project` dataclass. So far, it stores a custom `mypy_cmd` and a `pyright_cmd`. I've noticed that the `pyright_cmd` in particular only comes in three variants:
- `pyright_cmd=None` — which means that Pyright shouldn't run for this project
- `pyright_cmd="{pyright}"` — which means that Pyright should be executed without passing any explicit paths
- `pyright_cmd="{pyright} some relative paths"` — which means that the command line includes a list of explicit paths that are passed to Pyright.
In particular, there are no `pyright_cmd` fields that pass additional flags to pyright.

For running other type checkers, it would be beneficial to have that information in a more structured form. This way we could reuse the list of explicit paths. So I'm proposing the following change:

* For projects that previously used `pyright_cmd=None`, we add a new flag `pyright_skip=True`.
* For projects that previously used `pyright_cmd="{pyright}"`, we simply remove that `pyright_cmd` parameter. We consider that to be the new default.
* For projects that previously used `pyright_cmd="{pyright} some relative paths"`, we also remove the `pyright_cmd` parameter and add a new `paths=["some", "relative", "paths"]` field.

I have not attempted to change the representation for mypy, but I could do so if it seems desirable. I'd like to discuss the general approach first.

I tested this by running
```
mypy_primer --type-checker pyright
```

### DIFF
--- a/mypy_primer/main.py
+++ b/mypy_primer/main.py
@@ -118,7 +118,7 @@ def select_projects() -> list[Project]:
         if not (p.min_python_version and sys.version_info < p.min_python_version)
     )
     if ARGS.type_checker == "pyright":
-        project_iter = iter(p for p in project_iter if p.pyright_cmd is not None)
+        project_iter = iter(p for p in project_iter if not p.pyright_skip)
     if ARGS.project_selector:
         project_iter = iter(
             p for p in project_iter if re.search(ARGS.project_selector, p.location, flags=re.I)
--- a/mypy_primer/model.py
+++ b/mypy_primer/model.py
@@ -25,7 +25,12 @@ class Project:
     name_override: str | None = None
 
     mypy_cmd: str
-    pyright_cmd: str | None
+
+    # Whether or not to run pyright on this project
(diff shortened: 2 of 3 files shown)


hauntsaninja:
Sorry for the delay! This has run into some merge conflicts; I fixed up and took a slightly different approach in https://github.com/hauntsaninja/mypy_primer/pull/151. Let me know what you think!

PR #136, opened 2025-03-13
Add `scipy-stubs` as project

jorenham:
It's a pretty beefy stubs-only package with 50k+ lines of (valid) code, so I figured it might be useful to add it here :)


erictraut:
This is a great addition to mypy_primer. Thanks Joren!

JelleZijlstra (approved):

PR #137, opened 2025-03-14
add `numpy` to the `deps` of `scipy-stubs`

jorenham:
I left it out earlier because `scipy` already requires `numpy`.

But by explicitly adding `numpy` to `deps`, `scipy-stubs` will also be selected when `--known-dependency-selector=numpy` is passed, as e.g. `numpy` itself does.

### DIFF
--- a/mypy_primer/projects.py
+++ b/mypy_primer/projects.py
@@ -804,7 +804,7 @@ def get_projects() -> list[Project]:
             location="https://github.com/scipy/scipy-stubs",
             mypy_cmd="{mypy}",
             pyright_cmd="{pyright}",
-            deps=["scipy", "optype"],
+            deps=["scipy", "optype", "numpy"],
             expected_mypy_success=True,
             expected_pyright_success=True,
         ),


hauntsaninja (approved):
Thanks!

Looks like this was added in #136... When I run `mypy_primer -k scipy-stubs --debug`, I get `mypy: can't read file 'tests/**/*.py': No such file or directory`. So effectively when using mypy as the type checker, nothing is getting checked.

jorenham:
> When I run `mypy_primer -k scipy-stubs --debug`, I get `mypy: can't read file 'tests/**/*.py': No such file or directory`. So effectively when using mypy as the type checker, nothing is getting checked.

Hm weird... It's a classic case of "but it works on my machine".

I'm guessing I should remove this line then?:
https://github.com/scipy/scipy-stubs/blob/0586ab1dd8611c62bf8df0ab6f59ee9b26b1c0d6/pyproject.toml#L113

hauntsaninja:
Nope, now I get `mypy: error: Missing target module, package, files, or command.`

On the old scipy-stubs commit, I can repro on Linux as well. Any chance this is a basedmypy vs mypy difference?

hauntsaninja:
Looks to be working with your 486 PR and https://github.com/hauntsaninja/mypy_primer/pull/139

jorenham:
> Any chance this is a basedmypy vs mypy difference?

Yea, might be then.

> Looks to be working with your 486 PR and #139

Pfew, ok; thanks for looking into it!

hauntsaninja:
Great! Thanks for the addition

PR #138, opened 2025-03-14
Migrate to pyproject.toml

sharkdp:
I'm not sure if this is a desired change for this project. Please feel free to close this without comment if you think it's unnecessary. `mypy_primer` works fine without this.

The idea is to replace the `setup.py`-based configuration with a `pyproject.toml` file. The main motivating reasons for this PR are:

- Being able to just use `uv/poetry/pdm run mypy_primer`.
- To declare the development dependencies so that I can run them without having to install them locally
- Being able to run `test.sh` without having to install black/flake8/isort/mypy globally
- To lock the dependencies to make sure that every developer uses the same version of black/flake8/isort/mypy when working on this project

I verified that it is still possible to install `mypy_primer` using `pip`, like it's done in the mypy CI pipeline [here](https://github.com/python/mypy/blob/bbd7a6cf5886ac7162f0f1cce5c34e89c807262e/.github/workflows/mypy_primer.yml#L45-L46).

Disclaimer: I work for @astral-sh. It's certainly *not* my goal to promote `uv` with this PR. I saw that `mypy_primer` already uses `uv` internally, for its testing, and in the CI pipeline. So I thought it would be fine to double-down on that decision. This PR also adds a `uv.lock` file to make sure everyone uses the same versions of the dev tools. I can also remove it if a tool-specific lock file (while we wait for [PEP-751](https://peps.python.org/pep-0751/)) is not desirable for some reason.


hauntsaninja (approved):
Thank you! And yeah, using uv internally made mypy_primer much faster :-)

PR #140, opened 2025-03-30
Fix issues running mypy_primer on Windows

Andrej730:
1) Use `get_npm()` for `npm` commad as on Windows it tends to be installed as `npm.cmd` and not `npm.exe` and `subprocess` module requires full filename in that case, not just `npm`.

2) Use `node` explicitly in commands that execute .js files instead of relying on shebang.

3) Fix activating npm on Windows:
- ensure `source` used only if it's not Windows
- use `&&` instead `;` to support running it on Windows (on Windows `;` doesn't separate commands but instead second command is passed as an argument to the first one)

There's a little caveat about the last bullet point - was it intended to skip errors with `;`? Tested `&&` on Windows and mypy_primer works fine. But when I was testing `&&` on unix using pyright's workflow, I've found that it no longer works - it seems it was relying on silent errors like the one below (https://github.com/Andrej730/pyright/actions/runs/14158580085/job/39661016528). Which seems odd since it either means next line was executed not inside venv which either may lead to other issues or makes venv activate unnecessary.
```
2025-03-30T18:04:46.6002618Z > source /tmp/mypy_primer/projects/_pydantic_venv/bin/activate; /tmp/mypy_primer/old_pyright/pyright_to_test/packages/pyright/index.js (19.0s)
2025-03-30T18:04:46.6003740Z 	/bin/sh: 1: source: not found
```

@hauntsaninja can you please take a look?

### DIFF
--- a/mypy_primer/main.py
+++ b/mypy_primer/main.py
@@ -71,8 +71,8 @@ async def setup_new_and_old_pyright(
 
     if ctx.get().debug:
         (new_version, _), (old_version, _) = await asyncio.gather(
-            run([str(new_pyright), "--version"], output=True),
-            run([str(old_pyright), "--version"], output=True),
+            run(["node", str(new_pyright), "--version"], output=True),
+            run(["node", str(old_pyright), "--version"], output=True),
         )
         debug_print(f"{Style.BLUE}new pyright version: {new_version.stdout.strip()}{Style.RESET}")
         debug_print(f"{Style.BLUE}old pyright version: {old_version.stdout.strip()}{Style.RESET}")
--- a/mypy_primer/main.py
+++ b/mypy_primer/main.py
@@ -19,7 +19,7 @@ from mypy_primer.globals import ctx, parse_options_and_set_ctx
 from mypy_primer.model import Project, TypeCheckResult
 from mypy_primer.projects import get_projects
 from mypy_primer.type_checker import setup_mypy, setup_pyright, setup_typeshed
-from mypy_primer.utils import Style, debug_print, line_count, run, strip_colour_code
+from mypy_primer.utils import Style, debug_print, get_npm, line_count, run, strip_colour_code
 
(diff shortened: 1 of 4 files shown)


jorenham:
This coincidentally also seems to solve #141, so thanks!

jorenham:
And for what it's worth; I was also seeing the `/bin/sh: 1: source: not found` before this fix on linux. In https://github.com/hauntsaninja/mypy_primer/issues/141#issuecomment-2767054406, to be precise.

hauntsaninja:
Hmm actually maybe the && just makes pyright not run at all. Maybe we need `.` for posix sh compat or something

jorenham:
maybe `executable="/bin/bash"` could help

jorenham:
> Hmm actually maybe the && just makes pyright not run at all. Maybe we need `.` for posix sh compat or something

According to the comments on https://stackoverflow.com/a/51663133/1627479 the issue is `source`. So I suppose that both your `.` and the `/bin/bash/` solutions should work then.

hauntsaninja (approved):

Andrej730:
@hauntsaninja 

image

There's still a problem with `;` intead of `&&` as it doesn't work on Windows and resulting in mypy-primer never running pyright at all:

> * use `&&` instead `;` to support running it on Windows (on Windows `;` doesn't separate commands but instead second command is passed as an argument to the first one)

E.g. below is an example of code executed from cmd:
```batch
py -c "import sys; print(sys.argv)"; node c:\index.js
:: ['-c', 'node', 'c:\\index.js']
```

hauntsaninja:
Ah sorry, thanks for catching! I got a little spooked by the `&&` because we just silently would never have run anything at all in https://github.com/hauntsaninja/mypy_primer/pull/140#issuecomment-2779654916

Opened a PR to fix!

PR #143, opened 2025-04-05
Fix venv activation with sh

### LINKED ISSUE #141: pyright cannot find dependencies, requires `--pythonpath`
When I locally run 

```shell
uv run mypy_primer --type-checker pyright -k scipy-stubs --debug --clear
```

around ~37068 errors are reported. It's only the first two errors that are relevant:

```
/tmp/mypy_primer/projects/scipy-stubs/scipy-stubs/__init__.pyi
  /tmp/mypy_primer/projects/scipy-stubs/scipy-stubs/__init__.pyi:4:6 - error: Import "numpy" could not be resolved (reportMissingImports)
/tmp/mypy_primer/projects/scipy-stubs/scipy-stubs/_lib/_disjoint_set.pyi
  /tmp/mypy_primer/projects/scipy-stubs/scipy-stubs/_lib/_disjoint_set.pyi:5:8 - error: Import "optype" could not be resolved (reportMissingImports)
```

So it looks like pyright isn't able to find the (installed) `numpy` and `optype` dependencies.

It's easy to confirm this by changing https://github.com/hauntsaninja/mypy_primer/blob/ba3a8ed842213b8bc877ce9f5a21859f96616f4d/mypy_primer/projects.py#L806 to

```py
pyright_cmd="{pyright} --pythonpath=/tmp/mypy_primer/projects/_scipy-stubs_venv/bin/python",
```

If we runn the same `mypy_primer` command again, we'll now see 43 errors (caused by the missing `packaging` dependency, I'll submit a PR soon).

---

I'm guessing that `scipy-stubs` isn't the only project with this issue. I'm not sure what the best solution here would be, but hardcoding the `--pythonpath` probably isn't the way to go here.

hauntsaninja:
Fixes #141

### DIFF
--- a/mypy_primer/model.py
+++ b/mypy_primer/model.py
@@ -263,7 +263,7 @@ class Project:
             env["MYPY_PRIMER_PREPEND_PATH"] = str(prepend_path)
 
         pyright_cmd = self.get_pyright_cmd(pyright, additional_flags)
-        pyright_cmd = f"source {self.venv.activate}; {pyright_cmd}"
+        pyright_cmd = f"{self.venv.activate_cmd}; {pyright_cmd}"
         proc, runtime = await run(
             pyright_cmd,
             shell=True,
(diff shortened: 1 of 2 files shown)


hauntsaninja:
cc @jorenham

PR #146, opened 2025-04-08
add async_utils

mikeshardmind:
This started out with being offered for red-knot's use over in astral's discord for visibility on new generic syntax, but mypy may benefit from a repository which uses that and type alias statements as well.


mikeshardmind:
PR will have to be remade due to how github handles edits by maintainers (They go directly on the branch, rather than on a new ref) and the change being offered to both upstream and a fork at once.

sharkdp:
> PR will have to be remade due to how github handles edits by maintainers (They go directly on the branch, rather than on a new ref) and the change being offered to both upstream and a fork at once.

Oh, I am very sorry. I had no idea this would affect your upstream PR.

mikeshardmind:
@sharkdp All good, that was my fault, It's an interaction I was aware of and didn't think about. remaking the pr from my local tree wasn't an issue.

PR #151, opened 2025-04-18
Record list of paths to type check

hauntsaninja:
Supersedes #135

This PR should be a no-op, but I'll later go through the cases not touched here to make sure we have good coverage on `paths` for red knot

Using templating is a little bit more explicit / boilerplate-y than the original PR, but keeps customisation easy


hauntsaninja:
cc @sharkdp

sharkdp:
Thank you very much, both for accepting this change at all, and for improving it!

PR #155, opened 2025-04-19
Upstream basic support for red knot

hauntsaninja:
Adapted for recent changes I made to make more mypy_primer logic type checker agnostic


hauntsaninja:
cc @sharkdp

sharkdp:
> This is adapted from your branch.

I was planning to clean that up a bit before upstreaming it, so thank you for doing that!

> The main functional difference is that it's still an explicit opt-in for projects, happy to rework that however is most convenient.

Honestly, the most convenient way for me so far was to be able to run it on *all* projects, even if no explicit Red Knot configuration was present. This is why I wanted to add that generic `paths` field in the project configuration. We currently run mypy_primer in Red Knot's CI with a much larger set of projects than what would be possible on this branch. So if you want to keep it opt-in, I would probably need to add a `knot_cmd` field to almost all projects soon, otherwise we could not run from this repo directly.

Red Knot currently still crashes when checking some projects in mypy_primer's list (we hope to be able to run on all projects very soon). So for this reason, one could argue that it might be desirable to have an opt-in parameter, since it allows users to run `mypy_primer --type-checker knot` without having to add a `--project-selector`. On the other hand, I don't know why anyone would do that at the moment, except for Red Knot developers/contributors — and we would certainly like mypy_primer to be able to run on all projects, even if it's a project where Red Knot currently fails.

To summarize, I would prefer if we could remove that `if p.knot_cmd is not None` filter and then make `"{knot} check"` the default command, unless there are `paths`, in which case it should be `"{knot} check {paths}"`. We could still keep `Project.knot_cmd` as a potential way to override the command for a specific project. But for now, we wouldn't need to set it on any project.

If you're okay with that change, I'm happy to make the changes myself, of course.

> I was also a little confused about the `subdir` thing (mypy_primer just clones multiple copies of the project to differently named directories by default), so I removed that, let me know if I misunderstood something

Thanks. I think that might have been a remnant of some early experimentation I did.

> Also let me know if you'd like commit rights!

Thank you for the offer. That would obviously be great, because it would allow us to make changes to `get_knot_cmd`/`setup_knot` when needed (still through a PR process of course), which in turn would allow us to use this repository in our CI. I wouldn't use the commit rights for anything else. And still go through normal PR review rounds when making any changes to non-purely-red-knot-related things.

mypy_primer/projects.py
```diff
@@ -85,13 +85,15 @@ def get_projects() -> list[Project]:
             location="https://github.com/hauntsaninja/mypy_primer",
             mypy_cmd="{mypy} -p mypy_primer --strict",
             pyright_cmd="{pyright} {paths}",
+            knot_cmd="{knot} check {paths}",
```
sharkdp, line 88: We could remove these `knot_cmd` lines now.

sharkdp (approved):
Thank you. Confirming that I can now [run our CI job successfully](https://github.com/astral-sh/ruff/pull/17500) from this branch.

PR #157, opened 2025-04-22
Add 'strawberry' as a project

sharkdp:
When working on `@dataclass_transform` support, I found it helpful to test on [`strawberry`](https://github.com/strawberry-graphql/strawberry), as it makes [heavy use of the feature](https://github.com/search?q=repo%3Astrawberry-graphql%2Fstrawberry%20dataclass_transform&type=code).

The `projects.py` file has a comment on top ("repos with plugins") which lists `strawberry`. I'm not sure if that entry implies anything regarding using it as a project. But I had to set `--config-file=` for mypy (as was done for some other projects), because of some errors when trying to run with `strawberry`'s [`mypy.ini`](https://github.com/strawberry-graphql/strawberry/blob/main/mypy.ini) (related to `--install-types` and `--non-interactive`).

I tested using mypy, pyright, and Red Knot.

### DIFF
--- a/mypy_primer/projects.py
+++ b/mypy_primer/projects.py
@@ -1510,6 +1510,13 @@ def get_projects() -> list[Project]:
             deps=["packaging"],
             expected_success=("pyright",),
         ),
+        Project(
+            location="https://github.com/strawberry-graphql/strawberry",
+            mypy_cmd="{mypy} {paths} --config-file=",
+            pyright_cmd="{pyright} {paths}",
+            paths=["strawberry"],
+            deps=["graphql-core", "typing-extensions", "python-dateutil", "packaging"],
+        ),
     ]
     assert len(projects) == len({p.name for p in projects})
     for p in projects:
--- a/mypy_primer/projects.py
+++ b/mypy_primer/projects.py
@@ -1515,7 +1515,7 @@ def get_projects() -> list[Project]:
             mypy_cmd="{mypy} {paths} --config-file=",
             pyright_cmd="{pyright} {paths}",
             paths=["strawberry"],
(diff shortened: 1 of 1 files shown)


mypy_primer/projects.py
```diff
@@ -1510,6 +1510,13 @@ def get_projects() -> list[Project]:
             deps=["packaging"],
             expected_success=("pyright",),
         ),
+        Project(
+            location="https://github.com/strawberry-graphql/strawberry",
+            mypy_cmd="{mypy} {paths} --config-file=",
+            pyright_cmd="{pyright} {paths}",
+            paths=["strawberry"],
+            deps=["graphql-core", "typing-extensions", "python-dateutil", "packaging"],
```
AlexWaygood, line 1518: It probably doesn't matter much since it's a light dependency that's quick to install, but I doubt we need to explicitly include `typing-extensions` as a dependency here. The stubs for `typing-extensions` are provided by typeshed (it pretends `typing-extensions` is part of the standard library so that we can e.g. use `typing-extensions` backports in `builtins.pyi` etc.)

```suggestion
            deps=["graphql-core", "python-dateutil", "packaging"],
```

sharkdp: It's listed as a dependency for a few other projects, so I wasn't sure...

hauntsaninja (approved):
Alex is right, but if I remember correctly even after the `--install-types --non-interactive` issue there was something fiddly with this plugin, so I'm happy to take this PR as is (it preserves the TODO at the top, which is good).

If strawberry is providing useful coverage for knot, it will also provide useful coverage for mypy without plugins

AlexWaygood (approved):

PR #159, opened 2025-04-23
Add knot_cmd to __repr__

stroxler:
I was poking at getting pyrefly working and noticed this was missing, I'm guessing it's just an omission from https://github.com/hauntsaninja/mypy_primer/commit/ebaa9fd27b51a278873b63676fd25490cec6823b


hauntsaninja (approved):
Thanks!

PR #160, opened 2025-04-24
Add pyrefly basic support

stroxler:
I'm just following the example from red-knot.

It seems to work, I developed using
```
mypy_primer -k janus --type-checker pyrefly --debug
```
and once all the crashes were fixed I got a handful of errors as expected.

I also ran
```
mypy_primer --type-checker pyrefly
```
which goes a long while and dumps quite a lot of errors since Pyrefly has a long way to go, but it seemed to run fine through to the end.

### DIFF
--- a/mypy_primer/model.py
+++ b/mypy_primer/model.py
@@ -28,6 +28,7 @@ class Project:
     mypy_cmd: str
     pyright_cmd: str | None
     knot_cmd: str | None = None  # TODO: remove this default
+    pyrefly_cmd: str | None = None  # TODO: remove this default
     paths: list[str] | None = None
 
     install_cmd: str | None = None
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -13,6 +13,10 @@ classifiers = [
     "Programming Language :: Python :: 3",
     "Topic :: Software Development",
 ]
+dependencies = [
+    "ipython>=8.35.0",
```
hauntsaninja, line 17: stray?

mypy_primer/model.py
```diff
@@ -344,6 +347,63 @@ async def run_knot(
...
+    ) -> TypeCheckResult:
+        env = os.environ.copy()
+        additional_flags = ctx.get().additional_flags.copy()
+
+        if typeshed_dir is not None:
+            additional_flags += ["--typeshed", quote_path(typeshed_dir)]
+        if prepend_path is not None:
+            env["MYPY_PRIMER_PREPEND_PATH"] = str(prepend_path)
+
+        env["CLICOLOR_FORCE"] = "1"
```
hauntsaninja, line 374: can remove if pyrefly doesn't use this env var

hauntsaninja (approved):
Looks good!

stroxler:
Addressed code review points.

I also realized the `--typeshed` flag was copypasta, Pyrefly currently cannot pick up a typeshed at runtime because the standard library stubs get injected directly into the binary

It should be possible to run mypy_primer on typeshed changes, but we have to inject the typeshedby overwriting the vendored typeshed at build time which seems like it should wait for a separate PR

PR #165, opened 2025-04-28
Don't crash primer on panic

hauntsaninja:
This way you don't have to look at job logs

### DIFF
--- a/mypy_primer/model.py
+++ b/mypy_primer/model.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import asyncio
 import difflib
 import os
+import random
 import re
 import shlex
 import shutil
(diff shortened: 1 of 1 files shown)


hauntsaninja:
cc @AlexWaygood would this a better workflow?

hauntsaninja:
Okay cool!

AlexWaygood:
(Partly this is because the comment workflow at ruff works slightly differently to the comment workflow at mypy/typeshed — rather than posting a new comment every time there's a new primer run, the bot edits an existing comment, which can make it hard to remember to check the comment that's way up near the top of the PR conversation thread. Personally I prefer mypy/typeshed's comment workflow, but the workflow we have at ruff currently is close to what folks on the team are used to from the ruff-ecosystem reports)

hauntsaninja:
I guess the diff nonce thing does make it such that it will continue to print in subsequent PRs

PR #169, opened 2025-05-14
Fix setup of static-frame

sharkdp:
`static-frame` project setup currently fails with:
```
      src/_arraykit.c:6:11: fatal error: numpy/arrayobject.h: No such file or directory
          6 | # include "numpy/arrayobject.h"
            |           ^~~~~~~~~~~~~~~~~~~~~
      compilation terminated.
      error: command '/usr/bin/cc' failed with exit code 1

      hint: This error likely indicates that you need to install a library that provides "numpy/arrayobject.h" for
      `arraykit@1.0.0`
```

I didn't look further into it, but arraykit [just released a new version](https://github.com/static-frame/arraykit/releases), so for now, pinning the dependency at 0.10.0 (the previous version) seems okay. If we decide to go with this, I can open an issue to track this, so we can eventually remove that pin again.

### DIFF
--- a/mypy_primer/projects.py
+++ b/mypy_primer/projects.py
@@ -1472,7 +1472,7 @@ def get_projects() -> list[Project]:
             mypy_cmd="{mypy} {paths}",
             pyright_cmd="{pyright} {paths}",
             paths=["static_frame"],
-            deps=["numpy", "arraykit"],
+            deps=["numpy", "arraykit==0.10.0"],
         ),
         Project(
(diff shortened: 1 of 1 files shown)


sharkdp:
The CI failure also reproduces on `master`. It seems unrelated to the change here.

AlexWaygood (approved):

hauntsaninja:
Yeah, the CI failure is because of a mypy change on master that I didn't anticipate, will fix

PR #170, opened 2025-05-15
Update costs to run mypy

A5rocks:
I ran mypy_primer with `--measure-project-runtimes` locally, turned the result into a csv with some shell commands, then attached it using `mypy_primer.projects.update_projects` after I mutated them to update costs. I readded the important formatting afterwards. Some things are obviously incorrect (e.g. `colour` is marked as taking 2 seconds and I'm fairly certain anything under 10 is probably somehow wrong. How did it get things wrong? I don't know.) but I wanted to put this up first before going through with a fine grained brush.

The approach described above should work for anyone who thinks they have a better setup than I did.

### DIFF
--- a/mypy_primer/model.py
+++ b/mypy_primer/model.py
@@ -55,8 +55,7 @@ class Project:
         result = f"Project(location={self.location!r}, mypy_cmd={self.mypy_cmd!r}"
         if self.name_override:
             result += f", name_override={self.name_override!r}"
-        if self.pyright_cmd:
-            result += f", pyright_cmd={self.pyright_cmd!r}"
+        result += f", pyright_cmd={self.pyright_cmd!r}"
         if self.ty_cmd:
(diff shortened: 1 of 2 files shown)


hauntsaninja (commented):
The balancing was originally done for typeshed, so I'd run the timings with mypyc-compiled mypy (and on a relatively fast machine, but the absolute number doesn't matter too much)

colour-science and zulip are definitely wrong though, that is because the setup was broken until a few months ago

A5rocks:
Oops sorry I meant to update the costs. But I figured out why mypy primer getting the runtime failed -- I guess mypy 1.15 failed on `colour`.

A5rocks:
Alright so I ran specifically `colour` and `zulip` with (uncompiled, but we can just scale): `zulip` was ~260 seconds both times, `colour` was ~2210 seconds. So I'll make a PR to update the cost for `colour` to ~800 soon, unless you make it first.

PR #171, opened 2025-05-19
psycopg: Configure src.root for ty

sharkdp:
As analyzed [here](https://github.com/astral-sh/ruff/pull/18137), psycopg uses a src layout but the src folder is called `psycopg`. This is not supported in ty, unless `src.root` is configured accordingly.


JelleZijlstra (approved):

hauntsaninja:
Yeah, sorry about that. More folks have been contributing to mypy_primer, so I wanted to have a test that actually runs mypy_primer, but I guess I was a little lazy/unimaginative... the projects it runs on are simple. Anyway, pinned the type checker versions now.

PR #186, opened 2025-06-26
Fix mypy primer for Pyrefly

yangdanny97:
A couple of adjustments:
- make the configuration logic match the other checkers
- don't show code snippets or summary for emitted errors
- fix the path for the built binary

Tested on https://github.com/facebook/pyrefly/pull/565

Unrelated: even though the "run pyrefly on mypy primer" job succeeds, I can't seem to get the "comment on the PR" job to trigger - do I need a special setup for that?

### DIFF
--- a/mypy_primer/main.py
+++ b/mypy_primer/main.py
@@ -50,12 +50,8 @@ def setup_type_checker(
         setup_fn = setup_ty
         kwargs = {"repo": ARGS.repo}
     elif ARGS.type_checker == "pyrefly":
-        return setup_pyrefly(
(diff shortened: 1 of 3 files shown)


hauntsaninja (approved):

hauntsaninja:
> Unrelated: even though the "run pyrefly on mypy primer" job succeeds, I can't seem to get the "comment on the PR" job to trigger - do I need a special setup for that?

I think you might need to merge your PR for the comment workflow to work. The comment workflow requires some amount of write perms, so I think you can't test it from the PR. (This is also why it is a separate workflow, since mypy_primer can run untrusted code)

PR #187, opened 2025-07-03
add missing `scipy-stubs` deps

jorenham:
This adds the scipy-stubs as dependency to the following projects:

- xarray (because https://github.com/pydata/xarray/blob/0074b763f231274a5197cff56e8609f4283cfcad/pyproject.toml#L54)
- freqtrade (because https://github.com/freqtrade/freqtrade/blob/87b0a6d3f302ad64097a8e84f30b83caeffdb629/pyproject.toml#L113)
- dedupe (see https://github.com/dedupeio/dedupe/pull/1225)
- apache spark / pyspark (because https://github.com/apache/spark/blob/e28f42700611f6acd2de7dc84ecdf4f5b86b8f71/dev/requirements.txt#L30)
- pandera (because https://github.com/unionai-oss/pandera/blob/f8384ae7a48c8ff6d3af1de7ce7539f1d1ea1c9b/pyproject.toml#L55)
- pandas-stubs (because https://github.com/pandas-dev/pandas-stubs/blob/82d1a659a0849f6001b3e11bf8e72b7a61b749d3/pyproject.toml#L64)
- static-frame (because https://github.com/static-frame/static-frame/blob/20ce568b647e98d7dee94a816e40403ff046f31a/requirements-test-3_13.txt#L8)


AlexWaygood (commented):
It _looks_ like the only one of these projects that actually installs scipy-stubs in CI before running type checkers on their code is optuna. I think in general we try to have mypy_primer emulate the experience these projects get in CI when they actually run type checkers on their code. That said, I can totally see that having scipy-stubs installed for these projects would make things easier for you to test scipy-stubs in your CI!

I wonder if you could do something with `MYPY_PATH` (and/or the equivalent pyright setting, if you're using it with pyright?) in your CI's invocation of mypy_primer so that mypy considers scipy-stubs installed for the purposes of your CI, but not for other users of mypy_primer...?

AlexWaygood:
(I'm pretty new as a maintainer here; curious if the other maintainers have any thoughts!)

hauntsaninja (commented):
I'm potentially okay with this, given that we're not currently installing scipy for these projects at all...

...but if your goal is to add mypy_primer to the CI of scipy-stubs, you'll need to do a path munging thing (and adding the dep here will not help with that, in fact it may even confuse)

You can see how we set this up in numpy CI, via the `--new-prepend-path` and `--old-prepend-path` flags. Let me know it that works!

jorenham:
> ...but if your goal is to add mypy_primer to the CI of scipy-stubs, you'll need to do a path munging thing (and adding the dep here will not help with that, in fact it may even confuse)
> 
> You can see how we set this up in numpy CI, via the `--new-prepend-path` and `--old-prepend-path` flags. Let me know it that works!

Thanks to your advice, I *think* I managed to get it to work now: https://github.com/scipy/scipy-stubs/pull/824

PR #189, opened 2025-07-07
Add `zope.interface`

AlexWaygood:
This project would have caught https://github.com/astral-sh/ty/issues/764 before the PR introducing the regression was merged

### DIFF
--- a/mypy_primer/projects.py
+++ b/mypy_primer/projects.py
@@ -1592,6 +1592,13 @@ def get_projects() -> list[Project]:
             cost={"mypy": 24},
         ),
         Project(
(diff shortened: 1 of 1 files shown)


AlexWaygood:
Here's the mypy-zope plugin: https://github.com/Shoobx/mypy-zope

I think if we enabled that while checking `zope.interface` itself, we might not actually check any of the code in `zope.interface` that currently makes ty crash (we might just check the stub files that come bundled with that plugin?). So while it might make sense to enable that plugin when type-checking `twisted`, I think it actually _doesn't_ make sense to enable it when type-checking `zope.interface` itself.

PR #205, opened 2025-09-14
Add flag to install `mypyc/lib-rt` alongside mypy

brianschubert:
c.f. https://github.com/python/mypy/issues/19844#issuecomment-3288922936

I put this behind a flag to avoid affecting other projects running mypy_primer that may not be prepared to compile an extension module in their CI

### DIFF
--- a/mypy_primer/globals.py
+++ b/mypy_primer/globals.py
@@ -17,6 +17,7 @@ class _Args:
     repo: str | None
     type_checker: str
     mypyc_compile_level: int | None
+    mypy_install_librt: bool
 
     custom_typeshed_repo: str
     new_typeshed: str | None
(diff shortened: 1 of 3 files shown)


mypy_primer/globals.py
```diff
@@ -97,6 +98,12 @@ def parse_options(argv: list[str]) -> _Args:
         type=int,
         help="Compile mypy with the given mypyc optimisation level",
     )
+    type_checker_group.add_argument(
+        "--mypy-install-librt",
+        action="store_true",
+        help="(Experimental) Whether to install the mypyc C runtime library "
```
A5rocks, line 104: IMO explicitly calling this experimental doesn't make sense. What, are we going to make a PR adjusting this language and then another PR removing this when it's standard?

A5rocks (approved):
This makes sense to me.

(To any readers who think "why can't we just tell pip to install only wheels", the issue is that `lib-rt` is not currently published on PyPI.)

PR #206, opened 2025-09-14
Temporarily drop `materialize` from projects

brianschubert:
See https://github.com/python/typeshed/pull/14720, https://github.com/python/typeshed/pull/14716#issuecomment-3289636151

`materialize` has VCS dependency on a GitHub repo that was recently deleted or made private. This is causing the venv setup to fail, and is resulting in mypy_primer hanging in various projects' CI.


AlexWaygood:
Thank you!

PR #209, opened 2025-09-16
use taskgroup to avoid cursed cpython bug

### LINKED ISSUE #207: Exceptions raised during project setup can result in mypy_primer hanging
See #206, https://github.com/python/typeshed/pull/14716#issuecomment-3289575766

Exceptions raised when installing project dependencies [here](https://github.com/hauntsaninja/mypy_primer/blob/cfea3ce5007c959dc78c67de138a06d66e3c8612/mypy_primer/model.py#L138) can somehow fail to be logged / re-raised, and can result in the process hanging.

hauntsaninja:
Fixes #207

https://github.com/python/cpython/issues/103847

### DIFF
--- a/mypy_primer/main.py
+++ b/mypy_primer/main.py
@@ -390,36 +390,37 @@ async def primer(ARGS: _Args) -> int:
         old_typeshed_dir=old_typeshed_dir,
     )
 
-    results = [
-        project.primer_result(
-            new_type_checker=new_type_checker,
-            old_type_checker=old_type_checker,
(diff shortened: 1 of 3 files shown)


AlexWaygood (approved):

sharkdp:
> Hmm, weird, I do see the print output locally... can you reproduce?

I could not reproduce this locally. Today, everything works fine in CI as well. I have no idea what happened. Sorry for the distraction.

PR #218, opened 2025-10-15
 Remove `--debug-build`, add `--cargo-profile` instead 

sharkdp:
* Remove the recently-introduced `--debug-build` option and add a `--cargo-profile` option instead. This allows for more flexibility (`--cargo-profile dev` corresponds to `--debug-build`)
* Make sure that pyrefly builds work for users with a custom `build.target-dir` setting by explicitly specifying the target-dir (same as for ty)
* Use `python-interpreter-path` for pyrefly runs (apparently, the name of the option has changed recently)

I tested this for both ty and pyrefly with different cargo profiles.

### DIFF
--- a/mypy_primer/model.py
+++ b/mypy_primer/model.py
@@ -351,7 +351,7 @@ class Project:
 
         pyrefly_cmd = pyrefly_cmd.format_map(_FormatMap(pyrefly=pyrefly, paths=self.paths))
 
-        pyrefly_cmd += f" --python-interpreter {quote_path(self.venv.dir)}/bin/python --summary=none --output-format min-text"
+        pyrefly_cmd += f" --python-interpreter-path {quote_path(self.venv.dir)}/bin/python --summary=none --output-format min-text"
         return pyrefly_cmd
 
     async def run_pyrefly(


mypy_primer/globals.py
```diff
@@ -144,7 +143,9 @@ def parse_options(argv: list[str]) -> _Args:
 
     proj_group = parser.add_argument_group("project selection")
     proj_group.add_argument(
-        "-k", "--project-selector", help="regex to filter projects (matches against location)"
+        "-k",
+        "--project-selector",
+        help="regex to filter projects (matches against location)",
```
sharkdp: Not sure why `black` decided to reformat this *now*...

AlexWaygood: Did you maybe add a trailing comma here accidentally...?

AlexWaygood (approved):

mypy_primer/globals.py
AlexWaygood: Let's change it back if it fits in the configured line length.

(I generally prefer to enable `skip-magic-trailing-comma = true` in my black configs so that you don't end up with surprises like this)

PR #224, opened 2025-11-30
add `numpy-stl`

jorenham:
https://github.com/wolph/numpy-stl

It's a pretty popular project, and passes mypy (strict), (based)pyright (strict), and pyrefly ([config](https://github.com/wolph/numpy-stl/blob/develop/pyproject.toml), [workflow](https://github.com/wolph/numpy-stl/blob/develop/.github/workflows/type_check.yml)), so I figured it'd make for a nice addition :)


jorenham:
cc @wolph

hauntsaninja:
Thank you!

PR #192, opened 2025-08-05
Retune costs for mypy

brianschubert:
Now that type checking colour is much faster (https://github.com/python/mypy/issues/19403#issuecomment-3148623379), the current shard distribution is a bit lopsided. In a recent run in mypy's CI, the fastest shard finished in ~8min while the slowest took ~20min.

This updates the mypy costs using timings measured on GitHub Actions runners  ([run here](https://github.com/brianschubert/mypy/actions/runs/16737266720)). I used interpreted mypy for these timings, but I could re-measure for mypyc-compiled mypy if that's preferred.

PR #226, opened 2025-12-01
TOML-based project definitions

jorenham:
In general it's good practice to separate the data sources from the code. It's also a bit more readable this way IMO.

But my main motivation was that I thought it would be cool if you could easily override/extend the current "default" list of projects with something like `--extend-projects=custom_projects.toml`. These changes should make it easy to implement a feature like that.

I verified that the resulting `list[Project]` match the previous one.

The `update_projects` function seemed like dead code to me, and with these changes wouldn't work anymore. Is it ok that I removed it? Or should I rewrite it to output a toml string or something?


A5rocks:
> The `update_projects` function seemed like dead code to me, and with these changes wouldn't work anymore. Is it ok that I removed it? Or should I rewrite it to output a toml string or something?

It's useful in case you want to (e.g.) update the costs. Past examples: https://github.com/hauntsaninja/mypy_primer/pull/192#issuecomment-3152968330, https://github.com/hauntsaninja/mypy_primer/pull/170#issue-3064890446. Anything that updates the TOML would be helpful and a good enough replacement.

jorenham:
> It's useful in case you want to (e.g.) update the costs. Past examples: [#192 (comment)](https://github.com/hauntsaninja/mypy_primer/pull/192#issuecomment-3152968330), [#170 (comment)](https://github.com/hauntsaninja/mypy_primer/pull/170#issue-3064890446). Anything that updates the TOML would be helpful and a good enough replacement.

Ok makes sense. It's just too bad that `tomllib` can't write to toml though, so we'd need to use something like [`tomli-w`](https://github.com/hukkin/tomli-w) for that. 

Would it be enough if I write an "inverse" `list[Project] -> dict[str, Any]` function? Then once it's needed, any method for converting the dict to toml could be used. That way we avoid having to choose and add an external dependency now.

A5rocks:
Actually maybe it would be nicer to have an example script (in some `scripts` folder) that updates costs from a CSV. That way we can use a TOML library that doesn't clobber the comments or spacing!

jorenham:
> Actually maybe it would be nicer to have an example script (in some `scripts` folder) that updates costs from a CSV. That way we can use a TOML library that doesn't clobber the comments or spacing!

Eh, CSV? I'm not sure I follow. 
But either way, it sounds like something that deserves its own PR then :thinking:

PR #233, opened 2026-01-07
Add textual to the archinstall deps

correctmost:
Reference:
https://github.com/archlinux/archinstall/blob/aabf6ae19ed526f77698d9b23ffccf240e7fc853/pyproject.toml#L24


hauntsaninja:
Thanks!

PR #235, opened 2026-01-09
Add PyCA cryptography to projects

### LINKED ISSUE #234: Consider adding cryptography
cryptography (a/k/a PyCA cryptography) is a top 20 package on PyPI and uses lots of type hints, both in public and internal APIs. I think it'd be a great addition here!

I'd be happy to send a PR adding it, if acceptable.

CC @Alex @reaperhulk

woodruffw:
I think I did this right, but please let me know if not! I'll leave some comments inline on things I wasn't sure about.

Closes #234.

### DIFF
--- a/mypy_primer/projects.py
+++ b/mypy_primer/projects.py
@@ -1697,7 +1697,7 @@ def get_projects() -> list[Project]:
             location="https://github.com/pyca/cryptography",
             mypy_cmd="{mypy} {paths}",
             pyright_cmd="{pyright} {paths}",
-            paths=["src/cryptography"],
+            paths=["src/cryptography", "tests"],
             deps=["cffi", "typing-extensions"],
             expected_success=("mypy", "pyright"),
             cost={"mypy": 6},


mypy_primer/projects.py
```diff
@@ -1693,6 +1693,15 @@ def get_projects() -> list[Project]:
             deps=["numpy", "python-utils"],
             expected_success=("mypy", "pyright", "pyrefly"),
         ),
+        Project(
+            location="https://github.com/pyca/cryptography",
+            mypy_cmd="{mypy} {paths}",
+            pyright_cmd="{pyright} {paths}",
+            paths=["src/cryptography"],
+            deps=["cffi", "typing-extensions"],
+            expected_success=("mypy", "pyright"),
```
woodruffw: I believe cryptography themselves only test against mypy, but I think pyright _should_ also pass.

mypy_primer/projects.py
```diff
@@ -1693,6 +1693,15 @@ def get_projects() -> list[Project]:
...
+            cost={"mypy": 6},
```
woodruffw: I wasn't sure what to put here, so I cargo-culted a value 😅 

From context this is a cost estimate for how long each type checker is expected to take, but I wasn't sure if that was relative or absolute over all projects.

PR #240, opened 2026-02-08
Remove `pkg_resources` from setuptools `paths`

charliermarsh:
## Summary

I believe `pkg_resources` was removed back in [May](https://github.com/pypa/setuptools/commit/8ba2f38), but that change just merged into the default branch [today](https://github.com/pypa/setuptools/commit/530d114).


AlexWaygood:
> I believe `pkg_resources` was removed back in [May](https://github.com/pypa/setuptools/commit/8ba2f38)

Hmmm, it looks like that PR has been open since May, but was only merged 5 hours ago!

AlexWaygood (approved):
Thanks!

PR #251, opened 2026-05-29
update project dependencies

AlexWaygood:
I had codex analyze all `unresolved-import` diagnostics from the most recent full run of ecosystem-analyzer, look at which ones were genuinely caused by missing imports in the mypy_primer project configurations, and add them to the project configurations. I told it to exclude any dependencies like pytorch that would take a long time to install, and also had it measure the runtime of ecosystem-analyzer (which depends on mypy_primer) before and after this PR, for good measure (the runtime of ecosystem-analyzer was basically unchanged).

As well as ^those changes, this PR also replaces all types-requests dependencies with requests (since types-requests is deprecated), and removes all typing-extensions dependencies (since typing_extensions stubs are part of typeshed).


JelleZijlstra:
I think when mypy-primer was created @hauntsaninja didn't use the project's actual declared dependencies for performance reasons. Is that still a good reason now that uv makes installation much faster? It would be nice if we don't have to maintain dependency lists here and can just install those projects.

hauntsaninja:
Yeah, performance is less of a concern now, although IIRC last I checked it can still matter (good to exclude pytorch!). The biggest issue today is `--known-dependency-selector` is load bearing in several projects' CI. (It's also not the worst to have a few more Any's, mypy_primer skews to well-typed code anyway)

AlexWaygood:
> Yeah, performance is less of a concern now, although IIRC last I checked it can still matter (good to exclude pytorch!).

Yup, I told codex to exclude pytorch and other expensive deps! And I checked that project setup time is not meaningfully impacted by all these new deps.

> It's also not the worst to have a few more Any's, mypy_primer skews to well-typed code anyway

True! I think there's still quite a few `Any`s from unresolved imports even with this PR, though, at least for ty. Codex reckons that this gets rid of 923 `unresolved-import` diagnostics from the ecosystem for ty, out of a total of 5,702. And at least for ty, I'd definitely prefer to have an ecosystem report that "makes sense" as much as possible and reflects as closely as possible what projects are most likely to see in CI when they actually run type checkers.

> I think when mypy-primer was created @hauntsaninja didn't use the project's actual declared dependencies for performance reasons. Is that still a good reason now that uv makes installation much faster? It would be nice if we don't have to maintain dependency lists here and can just install those projects.

Yeah, I'd definitely be happy for us to just use projects' own dependencies lists more rather than hardcoding this in mypy_primer! This PR doesn't really change the status quo there, though; we could still easily make that change in a followup if we wanted to. And I think even if we did use custom `install_cmd` commands (or something other way of using projects' own dependency lists) more, we might need to still hardcode deps for quite a few projects, since it's sort-of unpredictable where type-checking deps are going to be listed, and whether/when they'll actually be installed when installing the project. (pyproject.toml? A custom dependency group in pyproject.toml? setup.cfg? requirements.txt? .pre-commit-config.yaml?) Sounds like it might also make the `--known-dependency-selector` flag less useful too?

PR #255, opened 2026-07-13
Add more projects which use pydantic

zanieb:
While adding pydantic support to ty, we noticed that the ecosystem analysis was not capturing changes relevant for pydantic features, especially those used in "application"-style projects rather than libraries. Here, I've identified a few popular projects which would be useful to evaluate.

I selected these projects by discovering projects that depend on pydantic via a GitHub search then prioritized a subset based on popularity and breadth of Pydantic usage.

Diagnostic changes in these projects would be reported in the following pydantic-related ty pull requests:

- `langgenius/dify`
  - https://github.com/astral-sh/ruff/pull/26628
  - https://github.com/astral-sh/ruff/pull/26629
  - https://github.com/astral-sh/ruff/pull/26648
  - https://github.com/astral-sh/ruff/pull/26655
  - https://github.com/astral-sh/ruff/pull/26762
  - https://github.com/astral-sh/ruff/pull/26764

- `NVIDIA/SkillSpector`
  - https://github.com/astral-sh/ruff/pull/26598
  - https://github.com/astral-sh/ruff/pull/26632
  - https://github.com/astral-sh/ruff/pull/26764

- `PostHog/posthog`
  - https://github.com/astral-sh/ruff/pull/26648
  - https://github.com/astral-sh/ruff/pull/26762

- `getsentry/sentry`
  - https://github.com/astral-sh/ruff/pull/26762
  
With the current corpus, only https://github.com/astral-sh/ruff/pull/26762 reported any diagnostic changes.


mypy_primer/projects.py
```diff
@@ -674,6 +674,49 @@ def get_projects() -> list[Project]:
             expected_success=("mypy",),
             cost={"mypy": 29, "ty": 5},
         ),
+        Project(
+            location="https://github.com/langgenius/dify",
+            mypy_cmd="{mypy} {paths} --exclude-gitignore --exclude '(^|/)conftest\\.py$' --exclude 'tests/' --exclude 'migrations/' --check-untyped-defs --disable-error-code=import-untyped",
+            pyright_cmd="{pyright} {paths}",
+            ty_cmd="{ty} check {paths} --force-exclude --exclude 'api/tests/**' --exclude 'api/providers/**/tests/**' --exclude 'api/migrations/**'",
```
zanieb: These commands seem a bit dubious? I'll investigate that

mypy_primer/model.py
```diff
@@ -124,6 +134,25 @@ async def setup(self) -> None:
...
+                    cwd=repo_dir / self.uv_sync_path,
+                    env=env,
+                    output=True,
+                )
+            except subprocess.CalledProcessError as e:
+                if e.output:
+                    print(e.output)
+                if e.stderr:
+                    print(e.stderr)
+                raise RuntimeError(f"uv sync failed for {self.name}") from e
```
zanieb, line 159: This might be a nice addition, but shouldn't we consider this separately? Why isn't this just the `install_cmd`?

mypy_primer/projects.py
```diff
@@ -674,6 +674,49 @@ def get_projects() -> list[Project]:
...
+        Project(
+            location="https://github.com/PostHog/posthog",
+            mypy_cmd="{mypy} {paths}",
+            pyright_cmd="{pyright} {paths}",
+            paths=["posthog", "ee", "products", "services/llm-gateway"],
+            uv_sync_path=".",
+            needs_mypy_plugins=True,
+            cost={"ty": 11},
+            min_python_version=(3, 13, 13),
+            max_python_version=(3, 13, 14),
```
zanieb, line 707: This feels dubious, we shouldn't need patch level granularity

mypy_primer/projects.py
```diff
@@ -674,6 +674,49 @@ def get_projects() -> list[Project]:
...
+        ),
+        Project(
+            location="https://github.com/NVIDIA/SkillSpector",
+            mypy_cmd="{mypy} {paths}",
+            pyright_cmd="{pyright} {paths}",
+            paths=["src/skillspector", "contrib/batch_scan"],
+            install_cmd="{install} '.[mcp]'",
+            cost={"ty": 1},
+            min_python_version=(3, 12),
+            max_python_version=(3, 15),
```
zanieb, line 696: Why do we need this?

mypy_primer/projects.py
```diff
@@ -674,6 +674,49 @@ def get_projects() -> list[Project]:
...
             cost={"mypy": 29, "ty": 5},
         ),
+        Project(
+            location="https://github.com/langgenius/dify",
+            mypy_cmd="{mypy} {paths} --exclude-gitignore --exclude '(^|/)conftest\\.py$' --exclude 'tests/' --exclude 'migrations/' --check-untyped-defs --disable-error-code=import-untyped",
+            pyright_cmd="{pyright} {paths}",
+            ty_cmd="{ty} check {paths} --force-exclude --exclude 'api/tests/**' --exclude 'api/providers/**/tests/**' --exclude 'api/migrations/**'",
+            paths=["api"],
+            uv_sync_path="api",
+            cost={"ty": 2},
```
zanieb: Weren't you also looking into cost calculations for mypy too?

mypy_primer/projects.py
zanieb: Is this the standard pattern for excluding tests?

zaniebot: Not as a shared primer abstraction. For mypy this follows the existing project-specific `--exclude` pattern; ty additionally needs `--force-exclude` because `api` is passed explicitly. I tested the unexcluded scope and it added 4,721 diagnostics, including 174 additional unresolved imports, almost all from tests and migrations. Excluding those keeps this focused on application code. Pyright does not expose an equivalent CLI exclusion, while Pyrefly uses Dify's own configuration.

mypy_primer/model.py
zaniebot: `install_cmd` currently requires `{install}` and expands it to `pip install` or `uv pip install` against the primer environment. These applications are not conventionally pip-installable: Dify relies on uv workspace sources, while PostHog and Sentry use their pyprojects for dependency management rather than packaging. `uv sync --active` is the supported way to realize those environments. I agree the generic hook is orthogonal enough to consider separately; I added it here to avoid repeating export pipelines in each definition.

mypy_primer/projects.py
zaniebot: Agreed that patch-level compatibility is unusual. This comes directly from PostHog's `requires-python = "==3.13.13"` and lockfile: `uv sync --frozen` rejects every other 3.13 patch, so the range makes primer skip the project instead of failing setup. That said, teaching the generic project model patch-level bounds solely for this repository is probably not worth it; avoiding the locked sync or leaving PostHog out would be cleaner.

mypy_primer/projects.py
zaniebot: SkillSpector declares `requires-python = ">=3.12,<3.15"`. Without the upper bound, primer would select it on Python 3.15 and its normal `pip install '.[mcp]'` would fail. This mirrors the package metadata and only needs minor-version granularity.

mypy_primer/projects.py
zaniebot: Yes. I only remeasured ty after changing setup: Dify 1.70–1.87s, SkillSpector 0.17–0.19s, PostHog 9.49–10.34s, and Sentry 2.64–2.71s. The enabled mypy, Pyright, and Pyrefly projects still inherit the default cost of 5, so this is not complete for non-ty sharding. I should measure those before marking this ready.

mypy_primer/model.py
zanieb: What sources does Dify use?

r2tbcb9n2c-hash (approved):

r2tbcb9n2c-hash (approved):

PR #257, opened 2026-08-04
Update attrs and Bokeh project configurations

sharkdp:


mypy_primer/projects.py
```diff
@@ -1515,6 +1515,7 @@ def get_projects() -> list[Project]:
                 "xyzservices",
             ],
             cost={"pyright": 60, "mypy": 65, "ty": 3},
+            min_python_version=(3, 12),
```
sharkdp, line 1518: The project itself requires 3.12. Checking with 3.11 leads to many syntax errors.

AlexWaygood (approved):

